PR #54, opened 2018-02-20
Flat serialization

ScatteredInk:
This is a first pass at what I think the flat serialisation will look like.

A couple of points to think about:

- Do we need the `type` field in `entityStatementReference`(etc)  fields for validation purposes or can we just get rid of it?
- Thinking further ahead, should our `entity` object also have separate fields for the different things that can be owned/controlled/benefitted from? Our _current_ use-cases are for registered companies/trusts but that could expand later, so dividing this field may make sense - although it has a cost in terms of readability.


examples/flat-serialisation/flat-single-direct.json
```diff
@@ -0,0 +1,76 @@
...
+    {
+      "id": "57d881b9-a35c-413e-870f-39ab8a1ac9e2",
+      "version": "0.1",
+      "beneficialOwnershipStatements": [
+        {
+          "id": "fbfd0547-d0c6-4a00-b559-5c5e91c34f5c",
+          "entity": {
+            "type": "entityStatement",
+            "statementDate": "2017-11-18",
+            "id": "1dc0e987-5c57-4a1c-b3ad-61353b66a9b7"
```
timgdavies: We should consider using `statementID` rather than just `id` to make clear that this links to another **statement** rather than an identified entity. 

The use of `id` is important to making use of standard  flatten tool and merge tooling, but I think (a) this is less of an issue if we move towards a complete cross-referenced statement model, and (b) I would anticipate we could always set tools to have a configuration option for this.

schema/beneficial-ownership-statements.json
```diff
@@ -75,30 +83,27 @@
...
             },
             {
-              "$ref": "#/definitions/PersonStatementReference"
+              "properties": {
+                "person": {
+                  "$ref": "#/definitions/PersonStatement"
+                }
+              }
             }
           ]
```
timgdavies: We need to either include option of null party here also, or include field at same level as `interestedParty` for declaring why not interested party is known.

schema/beneficial-ownership-statements.json
```diff
@@ -57,49 +66,53 @@
...
-            }
+            { "required": [ "subjectEntity" ] }
           ]
         },
         "interestedParty": {
           "title":"Interested party",
           "description":"The interested party has some level of ownership or control over the entity referenced in this beneficial ownership statement.",
+          "type": "object",
+          "properties": {
+            "interestedEntity": {
```
timgdavies: we could just say 'entity', 'person' or 'unspecified' here, as the nature of the relationship is given above

examples/flat-serialisation/anonymous-owner.json
```diff
@@ -0,0 +1,50 @@
...
+              "id": "00876127"
+            }
+          ]
+        }
+      ],
+      "personStatements": [
+        {
+          "statementID": "3154a26d-d29e-4db0-8815-866f57c75f0c",
+          "statementDate": "2017-07-18",
+          "missingPersonType": "anonymousPerson",
```
timgdavies: 'Missing person' feels like it could be read the wrong way. 

Perhaps 'unspecifiedPerson'  or just 'personType':['anonymous','identified'] and something like 'redactionNote' if it would be legitimate to refer to this as redaction?

examples/flat-serialisation/anonymous-owner.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "statementID": "32e09b90-203d-42d1-bd77-a6465bca2552",
+      "beneficialOwnershipStatements": [
+        {
+          "statementID": "0e5c86db-f2b0-4e11-bff0-db9eee3db221",
+          "statementDate": "2017-07-18",
+          "subject": {
+            "subjectEntity": {
+              "type": "entityStatement",
+              "statementID": "77ee64ba-858e-4413-bfaf-991be2200f71"
```
timgdavies: I've been wondering about whether using 'describedByStatement' as the property name here can help communicate the idea that we're linking to a description of the entity / person, rather than to the identified thing itself?

examples/flat-serialisation/anonymous-owner.json
```diff
@@ -0,0 +1,50 @@
+{
+  "version": "0.1",
+  "statementGroups": [
+    {
+      "statementID": "32e09b90-203d-42d1-bd77-a6465bca2552",
+      "beneficialOwnershipStatements": [
+        {
+          "statementID": "0e5c86db-f2b0-4e11-bff0-db9eee3db221",
+          "statementDate": "2017-07-18",
+          "subject": {
```
timgdavies: Do we need the complex structure here? Why subject rather than entity?

ScatteredInk: This will have the same structure as interestedParty, with more than just entities as possible targets of possible beneficial ownership statements. We will need 'arrangement', at least, to describe trusts and joint ventures.

examples/flat-serialisation/anonymous-owner.json
ScatteredInk: Yes, good idea.

examples/flat-serialisation/anonymous-owner.json
ScatteredInk: Yes, I think there are two separate things that can be separated out here, in two different fields: the type of missing data we are dealing with (redacted, unknown, known)  and the reason the data is missing (company didn't file, company doesn't need to file, BO didn't respond, intermediary didn't respond, etc). I don't have that in schema form, yet, but it's next on my list.

ScatteredInk:
Closing - this work is now taking place in the very-flat branch and the structure has changed per #57 .

Issue #63, opened 2018-03-08
Prepare clear folder structure of test data: valid and invalid


robredpath:
- Decide on folder structure
- Create test data

Bjwebb:
I think this was done in https://github.com/openownership/data-standard/commit/24ea3ad9a7ecb3db30b7849565065ef57955b797

Issue #62, opened 2018-03-08
Refactor schema into multiple sub-schemas for validation


timgdavies:
We have recognised that for validation of a fully flat model, we need to:

* Detect the statement type;
* Validate with a statement specific schema;

This suggests we may want to restructure into sub-schemas for each statement type, and then an overall schema that pulls these together.

robredpath:
Met this morning to discuss this work. The validation work depends on this being at least started. Can we confirm:
- When this is likely to start, and when we can meet to discuss details
- That @timgdavies and @ScatteredInk are actually doing this work, and it's not waiting on the developers

Bjwebb:
I think this was done in https://github.com/openownership/data-standard/commit/24ea3ad9a7ecb3db30b7849565065ef57955b797

Issue #57, opened 2018-03-08
Resolve questions over implementation of multiple serialisations of BODS data


robredpath:
In our updated modelling, we anticipate up to three kinds of BODS serialisation.

**(1) Array of statements**

E.g. 

```json
[
  {
    "id": "e9eeaaa3-d4d0-4036-8f47-731c5ddbf6c8",
    "type": "personStatement",
    "name":"Ed Example"
  },
  {
    "id": "d9cf4a36-fe23-4842-9dbe-5766f8744a05",
    "type": "entityStatement",
    "name":"E.G Corporation"
  },
  {
    "id": "e1fe7861-0494-406d-82c3-670652a9544c",
    "type": "beneficialOwnershipStatement",
    "subject": {
      "entity": {
        "identifiedByStatementID": "d9cf4a36-fe23-4842-9dbe-5766f8744a05"
      }
    },
    "interestedParty":{
        "person":{
            "identifiedByStatementID":"e9eeaaa3-d4d0-4036-8f47-731c5ddbf6c8"
        }
    }
  }
]
```

**(2) Statement JSON lines**

I.e. each statement on a single line.

```jsonl
{"id": "e9eeaaa3-d4d0-4036-8f47-731c5ddbf6c8","type": "personStatement","name":"Ed Example"}
{"id": "d9cf4a36-fe23-4842-9dbe-5766f8744a05","type": "entityStatement", "name":"E.G Corporation"}
{"id": "e1fe7861-0494-406d-82c3-670652a9544c","type": "beneficialOwnershipStatement","subject": {"entity": {"identifiedByStatementID": "d9cf4a36-fe23-4842-9dbe-5766f8744a05"}},"interestedParty":{"person":{"identifiedByStatementID":"e9eeaaa3-d4d0-4036-8f47-731c5ddbf6c8"}}}
```

**(3) Nested API responses**

Provided as an array of beneficial ownership statements, where the person and entity statements are nested. The below is only a draft, as we may want to consider API standard approaches for this. 

This would be recommended **only** for interactive API responses.

```
[
  {
    "id": "e1fe7861-0494-406d-82c3-670652a9544c",
    "type": "beneficialOwnershipStatement",
    "subject": {
      "entity": {
        "id": "d9cf4a36-fe23-4842-9dbe-5766f8744a05",
        "type": "entityStatement",
        "name": "E.G Corporation"
      }
    },
    "interestedParty": {
      "person": {
[...]

Issue #56, opened 2018-03-08
Create a Jupyter notebook that gives pass/fail validation for BODS data


robredpath:
We need a way to validate BODS data. Initially, this should be a Jupyter notebook. Over time, this will need to be a cmdline tool and ultimately a web app. 

It doesn't need to run entirely automatically at first. It's ok if there are steps which have to be done manually. It just needs to go through the process. 

The initial users for this are @ScatteredInk and @kd-ods

robredpath:
Depends on #62 and folder structure to be finished. Needs some sample data as well. 

Initial work can start before they're done. 

Key outcome is learning for CoVE.

Bjwebb:
What’s done (most of this is in 0.1-dev, but there’s still a pull request outstanding https://github.com/openownership/data-standard/pull/66):
* Functions to validate a BODS statement or package
* Similar functions that iterate over all errors for a BODS statement or package
* The above include detecting the appropriate statement schema from `statementType`
* Checking that statements referenced by a beneficialOwnershipStatement exist previously in a file https://github.com/openownership/data-standard/issues/59
* Running tests of example and test data with py.test on Travis

What’s not done:
* Jupyter notebook - @ScatteredInk will create the notebook/scripts he needs from the functions I’ve written. I’m available to answer any queries about this.
* JSON lines - but, I feel comfortable this should be straightforward to add.
* Useful validation messages - the tests I’ve written only use [ValidationError](https://python-jsonschema.readthedocs.io/en/latest/errors/#jsonschema.exceptions.ValidationError)’s message attribute. Additionally, path information doesn’t specify which statement the error was in. This is probably a priority to fix in the next sprint.

Other notes for doing this in CoVE:
* BODS validation is different from 360/OCDS, but it should just be a matter of replacing jsonschema’s `iter_errors` method with our custom functions.
* Sets of ids are going to be majority of memory usage. If this becomes a problem we may need to write these to disk (using a sqlite db would be a fairly straightforward approach)

ScatteredInk:
I've put together a cleaner version of this now, using @Bjwebb's test functions, in #67 

The use case for the notebook that I see at the moment is primarily around writing and checking new example data, so this notebook is primarily about loading in JSON and then showing the validation errors.

At this point, I'm not sure we need much else - unless you have any thoughts @kd-ods @timgdavies ?

timgdavies:
You can take that two-year silence as no thoughts from us.. ;) 

[...]

PR #64, opened 2018-03-14
Refactored schema and sub-schema

ScatteredInk:
This covers:

- refactoring into multiple schemas #62
- a directory structure for test data (I've used separate directories for each sub-schema) #63 
- the start of a validation notebook (just what I've been using for validating with the jsonschema library) #56 

The serialisation approach is option 1 from #57.


examples/flat-serialisation/anonymous-owner.json
```diff
@@ -0,0 +1,50 @@
+{
```
timgdavies: I don't think this aligns with Option 1 from #57 so we should remove these from the file. 

Ideally the examples directory should include a full valid statement package.

examples/flat-serialisation/gb-coh-entity-statement.json
```diff
@@ -0,0 +1,13 @@
+{
+  "statementID": "1dc0e987-5c57-4a1c-b3ad-61353b66a9b7",
```
timgdavies: Missing `statementType`?

examples/flat-serialisation/gb-coh-person-statement.json
```diff
@@ -0,0 +1,26 @@
+{
+  "statementID": "019a93f1-e470-42e9-957b-03559861b2e2",
+  "statementDate": "2017-11-18",
```
timgdavies: Missing 'statementType`?

examples/flat-serialisation/no-notifiable-owner.json
```diff
@@ -0,0 +1,42 @@
+{
```
timgdavies: As above. This example should probably be removed as it does not match option 1 from #57

examples/flat-serialisation/single-direct.json
```diff
@@ -0,0 +1,79 @@
+{
```
timgdavies: As above. This example should probably be removed as it does not match option 1 from #57

schema/beneficial-ownership-statements.json
```diff
@@ -1,8 +1,15 @@
 {
```
timgdavies: For the time being, I think we should drop beneficial-ownership-statement**s**.json (plural) as this doesn't fit the model in #57.

PR #65, opened 2018-03-20
Move pytest tests out of the notebook and test with Travis

Bjwebb:


tests/test_data.py
```diff
@@ -0,0 +1,85 @@
...
+from bods_validate import this_dir, format_checker, absolute_path_to_schema_dir, resolver
+from bods_validate import MissingStatementTypeError
+from bods_validate import bods_validate_package, bods_validate_statement
+
+
+@pytest.mark.parametrize('json_path', [
+    'data/entity-statement/valid/valid-entity-statement.json',
+    'data/person-statement/valid/valid-person-statement.json',
+    'data/beneficial-ownership-statement/valid/valid-beneficial-ownership-statement.json',
+    pytest.mark.xfail('../examples/flat-serialisation/gb-coh-entity-statement.json'),
```
ScatteredInk: This example should be fixed in latest commit, so we'd expect it to pass now.

tests/test_data.py
```diff
@@ -0,0 +1,85 @@
...
+    '../examples/flat-serialisation/gb-coh-person-statement.json',
```
ScatteredInk: This example should also be fixed.

tests/test_data.py
```diff
@@ -0,0 +1,85 @@
...
+def test_valid_statement_json(json_path):
+    with open(os.path.join(this_dir, json_path)) as f:
+        json_data = json.load(f)
+    bods_validate_statement(json_data)
+
+
+@pytest.mark.parametrize('json_path', [
+    'data/bods-package/valid/valid-bods-package.json',
+    '../examples/flat-serialisation/gb-coh-bods-package.json',
+    pytest.mark.xfail('../examples/flat-serialisation/single-direct.json'),
```
ScatteredInk: This example has been removed in latest commit, pending an updated version.

tests/test_data.py
Bjwebb: I've rebased onto 0.1-dev, but this still seems to fail.

ScatteredInk: Ah, thanks, that should be fixed now.

Bjwebb:
Thanks @ScatteredInk, the xfails are gone now.

PR #67, opened 2018-03-22
Notebook

ScatteredInk:
@Bjwebb Could you take a look at this to check if the changes to the tests to match new files are what you are expecting?


Bjwebb:
Yes, this mostly looks good. BTW, I think you could replace:
```python
    try:
        bods_validate_statement(statement)
    except ValidationError as e:
        for error in bods_iter_errors_statement(statement):
            print(str(error))
```
with just
```python
    for error in bods_iter_errors_statement(statement):
        print(str(error))
```

ScatteredInk:
Made simplifications suggested by @Bjwebb and merging.

Issue #59, opened 2018-03-08
Ordering of statements and validation


timgdavies:
Validation of BODS data requires checking:

(a) That each statement is structurally valid;
(b) That all the statements referenced by a beneficialOwnershipStatement **exist**;

This is made a lot easier if we impose an ordering constraint on data that all the constituent statements of a beneficial ownership statement must appear **before** the beneficial ownership statement itself. 

JSON arrays are ordered, so if as per #57 we use [ ] as our top-level, or JSON lines (which are also sequentially ordered), then parsers would be able to enforce this constraint.

Bjwebb:
I've written a checker that statements referenced by a beneficialOwnershipStatement exist previously in a file: https://github.com/openownership/data-standard/pull/66/commits/ed908d73259979b97571f3d05df71040a282abbd

Bjwebb:
@timgdavies Does this look like it does what we want?

timgdavies:
I added [this extra test file](https://github.com/openownership/data-standard/pull/68/files#diff-4cc802c0b1d2ade278ef273db7a9ae69) to check that this is also picking up badly ordered, as well as missing, referenced statements. 

I think from the fact that passed travis tests, with a new file in the folder that should fail secondary validation, that means the test correctly fails this (not using no double negatives here or anything). 

If you are happy that the test correctly fails my new file - then yes - this looks good to close.

timgdavies:
The remaining task then for this issue will be for us to make sure this restriction is captured in the documentation

Bjwebb:
The test file looks good, and I've added it to the tests successfully (a manual step atm, but means we can test more specifically than just pass/fail).

ScatteredInk:
Our tests are permissive about extra fields at the moment.

Ideally, a statement like this would fail rather than pass: 

```
{
 "statementID":"123",
 "statementType": "entityStatement",
 "removedFieldName": "bad data",
 "currentFieldName": "good data" 
}
```

timgdavies:
The best I can test, given current shift to JSCC for testing (#307) it looks to me as though:

* Test data in the repo fails py.test if statements are mis-ordered, or if there are additional fields
* The [data review tool](https://datareview.openownership.org/data/703cde53-c5af-402f-9371-d26ccf148ff4) reports on ordering and additional fields

So - I'm considering this closed.

PR #68, opened 2018-03-27
Testing statement order

timgdavies:
In relation to #59, a file should also fail validation when:

* All the statements referenced by a beneficial ownership statement are present **but** the beneficial ownership statement occurs **before** the statements that it references.


Bjwebb:
Note that the tests don't pick up new files automatically. I've added the new file.

PR #73, opened 2018-04-18
59 interested party

ScatteredInk:
I've added two test files that cover:

- a valid BODS package where a legal entity has another legal entity as its interested party.
- an invalid BODS package where a legal entity has another legal entity as its interested party (but the interested party entity statement is missing from the package)

The existing test for valid IDs only checks the case where an interested party is a natural person. If an entity is an interested party, then `UnrecognisedStatementID` will be raised regardless of whether the entity is present or not.


tests/bods_validate.py
```diff
@@ -49,10 +49,15 @@ def check_ids(statement, entity_statement_ids, person_statement_ids):
...
-            raise UnrecognisedStatementID("interestedParty/person/describedByStatement '{}' does not match any known persons".format(person_statement_id))
+            raise UnrecognisedStatementID("subject/entity/describedByStatement '{}' does not match any known entities".format(entity_statement_id))
+        if 'person' in statement.get('interestedParty', {}).keys():
+            person_statement_id = statement.get('interestedParty', {}).get('person', {}).get('describedByStatement')
+            if person_statement_id not in person_statement_ids:
+                raise UnrecognisedStatementID("interestedParty/person/describedByStatement '{}' does not match any known persons".format(person_statement_id))
+        if 'entity' in statement.get('interestedParty', {}).keys():
+            interested_entity_statement_id = statement.get('interestedParty', {}).get('entity', {}).get('describedByStatement')
+            if interested_entity_statement_id not in entity_statement_ids:
+                raise UnrecognisedStatementID("subject/entity/describedByStatement '{}' does not match any known entities".format(interested_entity_statement_id))
```
Bjwebb: `subject/entity/describedByStatement` should be `interestedParty/entity/describedByStatement`

tests/bods_validate.py
```diff
@@ -49,10 +49,15 @@ def check_ids(statement, entity_statement_ids, person_statement_ids):
     if statement_type == 'beneficialOwnershipStatement':
         entity_statement_id = statement.get('subject', {}).get('entity', {}).get('describedByStatement')
         if entity_statement_id not in entity_statement_ids:
-            raise UnrecognisedStatementID("subject/entitiy/describedByStatement '{}' does not match any known entities".format(entity_statement_id))
-        person_statement_id = statement.get('interestedParty', {}).get('person', {}).get('describedByStatement')
-        if person_statement_id not in person_statement_ids:
-            raise UnrecognisedStatementID("interestedParty/person/describedByStatement '{}' does not match any known persons".format(person_statement_id))
+            raise UnrecognisedStatementID("subject/entity/describedByStatement '{}' does not match any known entities".format(entity_statement_id))
+        if 'person' in statement.get('interestedParty', {}).keys():
```
Bjwebb: `.keys()` isn't necessary, and I think people argue that therefore you shouldn't use it here, but my google fu is failing me atm.

ScatteredInk:
> Is it possible to have both interestedParty/person and interestedParty/entity? Should we raise an error in that case?

No. There should only be one interestedParty. The schema should catch this, so I'm not sure if we need to test for this at the package level?

I added an invalid example file but now realise that my test was relying on Python 3.6' dictionaries being ordered, and our Travis is using Python 3.5, so what always works on my machine fails half the time here. Is there a better strategy for testing error messages that could be either A or B that we should take?

ScatteredInk:
I've rolled back the failing tests I've created and moved into a new branch for next validation targets, so I think this can be merged for now.

PR #74, opened 2018-04-18
45 loose validation

ScatteredInk:
To accommodate existing sources of data, we will need a method of relaxing validation for certain fields that, in the beta release, were either validated by the schema or intended to be validated using secondary methods. This includes:

- `nationalities` - for describing nationalities held by a natural person using ISO 2-Digit country code;
- `jurisdiction` - the ISO_3166-2 2-Digit country code, or ISO_3166-2 sub-division code of the responsible jurisdiction of an entity; and,
- `scheme` in the `Identifier` object. This uniquely identifies a register for a identification number, using org-id lists.

This PR is a first pass at representing loosely validated data on jurisdictions and scheme.

The [example data](https://github.com/openownership/data-standard/blob/2cc213be3d6b95bf67fcf0641f3fccd8f6738dbf/tests/data/bods-package/valid/valid-bods-package-loose-validation.json) is taken from this [company](https://beta.companieshouse.gov.uk/company/03209262/persons-with-significant-control) in the UK PSC register. Strict validation would remove a lot of the information contained in the declaration, even if it is a signal, rather than a definite statement of the legal owner of the company.

### DIFF
--- a/examples/flat-serialisation/gb-coh-bods-package.json
+++ b/examples/flat-serialisation/gb-coh-bods-package.json
@@ -10,7 +10,8 @@
     "identifiers":[
       {
         "scheme":"GB-COH",
-        "id":"07444723"
+        "id":"07444723",
+        "validationType": "strict"
       }
     ]
   },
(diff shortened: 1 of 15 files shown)


ScatteredInk:
Continuing in #80

PR #75, opened 2018-04-23
Primer comments

ScatteredInk:
A couple of clarifications/hedges on the documentation.

### DIFF
--- a/docs/whatisbo.md
+++ b/docs/whatisbo.md
@@ -10,17 +10,17 @@ The idea that people own companies and sometimes companies own companies is well
 
 ## ...but it's people that ultimately benefit...
 
-Even when companies own companies, individuals always appear at the end of the ownership chain. They ultimately benefit from companies’ financial successes, and often share in their failures.
+Even when companies own companies, individuals almost always appear at the end of the ownership chain. They ultimately benefit from companies’ financial successes, and often share in their failures.
 
 ![Company 2 owns 100% of Company 3's shares. And Person B owns 30% of Company 2's shares](_assets/Diag2-simpleInd.svg)
 
 ## ...so, we need to know who ultimately controls a company...
 
-If a person owns something, they usually have some control over it. Owning shares in a company may confer voting rights, for example. But ownership and control are sometimes separated - by contracts, agreements and other mechanisms. For example, an executive director might have the right to appoint 60% of the other board members.
+If a person owns something, they usually have some control over it. Owning shares in a company may confer voting rights, for example. But ownership and control are sometimes separated - by share classes, contracts, agreements and other mechanisms. For example, an executive director might have the right to appoint 60% of the other board members.
 
 ![Person B owns 30% of Company 2's shares. And Person C (an executive director) has the right to appoint 60% of Company 2's board members](_assets/Diag3-splitContr.svg)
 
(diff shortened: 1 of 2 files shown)


docs/whatisbo.md
```diff
@@ -10,17 +10,17 @@ The idea that people own companies and sometimes companies own companies is well
 
 ## ...but it's people that ultimately benefit...
 
-Even when companies own companies, individuals always appear at the end of the ownership chain. They ultimately benefit from companies’ financial successes, and often share in their failures.
+Even when companies own companies, individuals almost always appear at the end of the ownership chain. They ultimately benefit from companies’ financial successes, and often share in their failures.
```
ScatteredInk: There some instances where a BO chain won't resolve to an individual (an SOE or edge cases like the legal office of the pope). We might want to think about a glossary-like mechanism for pointing people towards "here's why we used the word 'almost' here.

docs/whatisbo.md
```diff
@@ -10,17 +10,17 @@ The idea that people own companies and sometimes companies own companies is well
...
 
 ![Company 2 owns 100% of Company 3's shares. And Person B owns 30% of Company 2's shares](_assets/Diag2-simpleInd.svg)
 
 ## ...so, we need to know who ultimately controls a company...
 
-If a person owns something, they usually have some control over it. Owning shares in a company may confer voting rights, for example. But ownership and control are sometimes separated - by contracts, agreements and other mechanisms. For example, an executive director might have the right to appoint 60% of the other board members.
+If a person owns something, they usually have some control over it. Owning shares in a company may confer voting rights, for example. But ownership and control are sometimes separated - by share classes, contracts, agreements and other mechanisms. For example, an executive director might have the right to appoint 60% of the other board members.
```
ScatteredInk: Shares with different voting rights is probably the most common way that ownership and control get separated.

docs/whatisbo.md
```diff
@@ -10,17 +10,17 @@ The idea that people own companies and sometimes companies own companies is well
...
 
 ![Person B owns 30% of Company 2's shares. And Person C (an executive director) has the right to appoint 60% of Company 2's board members](_assets/Diag3-splitContr.svg)
 
-Significant control over a company’s composition and decisions may be used to steer benefits and direct risks. So we need to know who has that control.
+An individual with significant control over a company’s board-composition and decision-making may be able to separate the risks undertaken by owners from the financial benefits derived from the company. So we need to know who has that control.
```
ScatteredInk: I've tried to get this towards the principle-agent problem that managers are taking risks with other people's money, which is what I think this is getting at. We may want to just reference that, I don't know.

docs/whatisbodata.md
```diff
@@ -7,7 +7,7 @@ Data relating to beneficial ownership may include:
 * Details of individuals
 * Details of other legal entities
 * A breakdown of share holdings 
-* Rules about voting rights
+* The voting rights attached to different share classes
```
ScatteredInk: Emphasise share classes.

docs/whatisbo.md
kd-ods: Indeed. 'The Zuckerberg hold'.

docs/whatisbo.md
kd-ods: You're right. We need to get that in there.

PR #80, opened 2018-05-11
loose validation

ScatteredInk:
This PR tries to reconcile the problem of distinguishing between validated and unvalidated beneficial ownership data.

We want to encourage best practice, while retaining the information from sub-optimal data (existing systems or paper-based systems) and keeping the analysis tasks manageable. The primary use-case here is reconciling entities and individuals using jurisdictions, nationalities and identifiers.

I have replaced the `jurisdiction` and `nationalities` string/array of string attributes with an object each for jurisdiction and nationality. My suggested guidance for using this is:

> A jurisdiction/nationality MUST have one of:
> 
> `code`: an ISO 2-digit (or 3-digit for jurisdiction) code;
> `name`: the name of the jurisdiction;
> `code` and `name`
> 
> Where BOTH a code and a name are given, the assumption is that the name overrides the code. i.e., an entity registered in the Dubai International Financial Centre might have the jurisdiction code "AE-DU" and the jurisdiction name "Dubai Financial Services Authority", reflecting the delegation of legal responsibility to this body.  

The other aspect of this PR is identifiers. Here, there is a choice between:

- preserving the purity of the `Identifier` object, i.e., it should just be a scheme and an id.
- putting the unvalidated information about where this company might be registered away from the Identifier object.

In this PR, I've gone for the former approach, putting uncertain information in an `incorporatedIn` attribute with jurisdiction.    

```
{
   "name":"Ibrakom Fzco",
    "incorporatedIn":{
      "jurisdiction": {
        "name": "Uae"
      },
      "companyRegisterName": "Jebel Ali Free Zone"
      },
    "identifiers":[
      {
        "id":"00335"
      }
    ]
}
```

An alternative way to represent this might be to represent the scheme as an object with a code (the org-id code) and an optional name (if the org-id code is unknown or the register is only provided as a string):

```
    "identifiers":[
      {
        "scheme": {
          "name": "Jebel Ali Free Zone"
        }
        "id":"00335"
      }
    ]
```


timgdavies:
I agree with the move to make jurisdiction an object with `name` and `code`, and to document the business logic applied here.

For identifiers, I would prefer an identifier block with:

* `scheme`
* `schemeName`
* `id`

all on the same level, as this maintains consistency with `identifier` blocks we have in other standards. 

Nesting wise, I wonder if we would be better to just have:

```
{
"incorporatedInJurisdiction":{
   "name":"UAE",
   "code":"UA"
},
"incorporatedInCompanyRegister":"Jebel Ali Free Zone"
}
```

where `incorporatedInCompanyRegister` is a property we would only ever want to see in loosely validated data, as in general there should be an `identifier/scheme` - and where we can seek to improve our `identifier/scheme` org-id codelist so that it's possible to validate for cases where someone tries to provide more than one official register.

ScatteredInk:
Thanks @timgdavies - I agree with the comments on nesting and have made the suggested changes. I've switched the base to `60-refactor-docs`.

I'm not sure when you are suggesting we use `schemeName` over, or as well as, `incorporatedInCompanyRegister` - can you clarify?

If `incorporatedInCompanyRegister` only appears in unvalidated data, then we have this as one option:

```
{
    "incorporatedInCompanyRegister":"Jebel Ali Free Zone",
    "identifiers":[
      {
        "id":"00335"
      }
    ]
  }
```

We know this company has declared a BO with company number 00335 in register Jebel Ali Free Zone - we also know that this information is unvalidated. 

We can potentially use `schemeName` to get the same information with a nicer nesting structure, if we say this is an unvalidated field.

```
{
    "identifiers":[
      {
        "id":"00335",
        "schemeName":"Jebel Ali Free Zone",
      }
    ]
  }
```

PR #86, opened 2018-05-29
Suggested primer updates

timgdavies:
This PR contains:

* A minor configuration update so we can use [all the Open Data Services Sphinx directives](https://sphinxcontrib-opendataservices.readthedocs.io/en/latest/misc/);
* Some suggested updates to phrasing, mostly in the primer;

One thing amongst these changes is a stylistic tweak, which I don't mind seeing reverted if Kadie/Jack have strong opinions otherwise on it: this change was to prefer referring to 'BODS' over 'the Standard', as the capitalisation of 'Standard' and 'Data Standard' didn't feel quite right to me. 

@kd-ods If you got chance to look at this PR before on leave and just let me know if anything you would object to heavily, that would be much appreciated, as I'll then also try some refactoring in terms of structuring markdown files in folders just for easier maintenance and governance.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,7 +31,7 @@ from recommonmark.parser import CommonMarkParser
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinxcontrib.jsonschema']
+extensions = ['sphinxcontrib.jsonschema','sphinxcontrib.opendataservices']
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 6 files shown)


kd-ods:
@timgdavies  - It's really great to have your eyes on this: all those changes are fine and welcome. Thanks! Yes, restructuring the files would be welcome too: I haven't had time to check how to do that, yet. I'll merge the pull request now.

(On the 'BODS' vs 'Standard' wording. I'm not wedded to the latter, I just wanted to give an acronym-free approach a chance. Let's use your approach for now and get @ScatteredInk's thoughts on it when he has a moment.)

PR #89, opened 2018-06-13
Start on front page - structure with placeholder text for others to fill in.

odscjames:
Converts .md to .rst - this is to get around big problems with .md. This may be changed back later, but we needed to progress this now.

At the moment everything appears in one column - but if this is accepted and merged into 60-refactor-docs that would allow @ScatteredInk and @kd-ods to work on the content and me to work on the style separately.

https://github.com/openownership/data-standard-sphinx-theme/issues/24


kd-ods:
Related IRC chat:
<Kadie> jacklord, james- I'm happy to approve the above PR. I just want timdavies to be aware of this pragmatic move to RST from markdown for the main index page (so that we can get this nice layout we want using divs). In case Tim knows of a good reason not to go in this direction.
<james> Good point, tho it may only be temporary but indeed timdavies
<timdavies> kadie james pragmatic is my middle name....  so merge away :)
<Kadie> timdavies, thanks :-)
<timdavies> kadie james And even long-term, I don't think we have to worry about a bit of mixing Markdown and RST - the key of why we're using Markdown for major bits of prose is just for easier write-ability of extended sections of documentation - but then RST for certain pages where it provides the best features seems good to me. when keeping for long-term, just think about our future selves coming to edit it, and if we need doc to remind us that's why! :)
<Kadie> timdavies, Pragmatic indeed! Yes - documenting the rationale is key.

PR #100, opened 2018-07-04
Convert pages from md to rst.

odscjames:
To get around a markdown problem with alt text appearing on the page or having the wrong alt tags.

Converted using pandoc

    pandoc data-standard/docs/primer/whatisbodata.md -f markdown  -t rst  -o  data-standard/docs/primer/whatisbodata.rst

and then manual tweaks to get rid of eval tags that now aren't needed, and indent text in alt tags properly.


timgdavies:
Great stuff @odscjames  - I've tested and this seems to be working. Merging now. 

There will be ways we can tidy up the rst a bit later, but this is working for now.

PR #109, opened 2018-08-03
Strict validation of BODS test data

odscjames:
http://bods.opendataservices.coop/redmine/issues/145


ScatteredInk:
Thanks James - looks good.

PR #126, opened 2018-11-14
Updates schema metadata to 0.2-beta: closes #125

### LINKED ISSUE #125: Update version in metadata of all statement types
- [x] Bods Package

- [x] Components

- [x] Entity Statement

- [x] Ownership or Control Statement

- [x] Person Statement

ScatteredInk:


ScatteredInk:
This will need to be updated for the release, depending on versioning convention - but at least means everything is in line for now.

ScatteredInk:
Updated this. We also:

- aren't using `id` to point to a uri.
- don't have a way to point to the version in the data itself.

I suggest we look at this in the next sprint in the context of compiled versions of the schema and validation tooling

PR #143, opened 2019-02-01
Sphinx and transifex config for translations

rhiaro:


README.md
```diff
@@ -26,3 +26,34 @@ The initial development of the Beneficial Ownership Data Standard is funded thro
...
+```
+$ pip install -r requirements.txt 
+```
+
+And you need to get a [Transifex API key](https://www.transifex.com/user/settings/api/), make sure you have access to the [BODS project on Transifex](https://www.transifex.com/OpenDataServices/bods-v01)
+
+**When you add or update the docs** you need to do the following so that they can be translated, in the `docs` directory:
+
+1. Run `make gettext` to extract translatable English strings.
+2. Run `sphinx-intl update -p _build/gettext -l fr -l ru` to generate translation (.po) files for the languages you need (pass languages with the `-l` flag).
```
Bjwebb: I'm not sure whether this step is necessary if you're using Transifex to do the translations.

PR #152, opened 2019-03-07
Set up redirects from pages that used to exist to pages that now do exist

odscjames:
This will make hopefully make browsing between versions on RTD nicer, in that you won't see 404 pages.


odscjames:
It is the master branch I should be using right?

Let me double check this works on RTD before this is reviewed.

odscjames:
This has now been tested for RTD and is ready for review

PR #155, opened 2019-03-20
Tax residency field added - fixes #107

### LINKED ISSUE #107: Tax residency
Tax residency would be a useful addition for person statements. This would add to OECD CRS compatibility and add to tax-based use-cases.

We could simply re-use the structure that we have for nationality:

```
    "taxResidencies": {
      "title": "Tax Residency",
      "description": "An array of ISO 2-Digit country codes representing the tax residencies held by this individual.",
      "type": "array",
      "items": {
        "$ref": "components.json#/definitions/Country"
      }
}
```

kd-ods:
Tax residency field added to person-statement.json. We'll add related examples later

### DIFF
--- a/examples/os-01-dr-01.json
+++ b/examples/os-01-dr-01.json
@@ -1,0 +1,76 @@
+[
+  {
+    "statementID": "a7b06b12-f5ff-49ff-a83f-41b2883d229b",
+    "statementType": "entityStatement",
+    "statementDate": "2019-05-15",
+    "entityType": "registeredEntity",
+    "name": "Company A",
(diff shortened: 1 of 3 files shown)


ScatteredInk:
This looks good - thanks @kd-ods and @siwhitehouse. We agreed to add example data to this PR before merging this in, so will leave this open until next week.

ScatteredInk:
I've added an example of a natural person with one declared nationality and two declared nationalities. Once this is agreed, I think we can close this.

kd-ods:
@ScatteredInk - that looks good to me.

Issue #10, opened 2017-01-23
Types of beneficial ownership relation


timgdavies:
> An issues related to the [draft conceptual model](https://github.com/openownership/data-standard/issues/7)

We will need a codelist to represent the different kinds of beneficial ownership and control relationships between an interested party, and the entity it controls.

This may need to be a hierarchical codelist, in which we have a number of top-level concepts, and then a range of sub-types. In the examples above, we have identified:

* Right to profits

    * Share ownership

* Right to direct decisions

    * Role-based

    * Contractual

* Role holder - with no implication of ownership or control (such as when a lawyer holds the role of being the nominee, but can only exercise that on behalf of another party)

We will need to find, or develop, a more detailed codelist here. 

Local implementations of the standard may need to map between this interoperable codelist and their own local legal requirements and systems.

sebbacon:
"Family member" is a kind of control often mentioned in legislation.

jpmckinney:
Re: local implementations mapping the codelist to local requirements, is there something we can learn from other standards? I know countries have mapped SIC, etc. that we might learn from.

ScatteredInk:
The UK PSC Register has a sub-category for LLPs that we can't represent with the current codelist:

1. Decisions
   * ...
2. Profits
   *  ...
   * Rights to surplus assets on winding up

ScatteredInk:
Extracting from #92, which I will close. We should add these control types - with better descriptions.

Interest | Description
-- | --
contract | A contract that defines beneficial ownership rights beyond those otherwise implied by ownership structures.
conditionalContract | An agreement that defines conditions by which beneficial ownership rights could come into existence

ScatteredInk:
Having read #178, I suggest renaming `rights-to-surplus-assets` to `rights-to-surplus-assets-on-dissolution` to properly reflect that control type.

timgdavies:
There are outstanding issues in version 0.2 related to:

(a) The description of [`interestType` codelist entries](http://standard.openownership.org/en/0.2.0/schema/reference.html#interesttype) discussed here https://github.com/openownership/data-standard/issues/178#issuecomment-497590414 . In particular I note that the description of `rights-granted-by-contract` and `conditional-rights-granted-by-contract` in the standard do not appear to reflect those in the issue.

(b) The conceptual distinction between roles vs. interests discussed above and here https://github.com/openownership/data-standard/issues/178#issuecomment-500401855 

As a result, we need to:

[...]

PR #184, opened 2019-05-22
add contractual rights, rename rights-to-surplus-assets-on-dissolution

ScatteredInk:
Adds contractual rights per #10, and renames `rights-to-surplus-assets` to `rights-to-surplus-assets-on-dissolution`, which more closely matches the definition in the UK's PSC register for LLPs.


ScatteredInk:
I'll leave this until #194 is merged - after which I don't think there will be anything to do.

ScatteredInk:
This has been done in #194

PR #189, opened 2019-05-23
Use translated schema and codelists in the docs

rhiaro:
Runs the translate process as part of the sphinx build, creates the translated schema and codelists into `docs/_build_schema` and updates references to json and csv files in the RST docs themselves to point to the right places.

(Adding @kd-ods @ScatteredInk and @siwhitehouse specifically to note a change to the contents of `reference.rst`.)


docs/conf.py
```diff
@@ -260,6 +259,8 @@ def translate_schema_and_codelists(language='en'):
         (glob(str(codelist_source_dir / '*.csv')), codelist_target_dir, codelist_domain),
     ], localedir, language, version=os.environ.get('TRAVIS_BRANCH', 'latest'))
 
+    print("Translated schema and codelists to {}".format(language))
```
odscjames: Did you mean to leave this in? I'm not necessary against some prints if it helps people running it, just checking

odscjames:
https://readthedocs.org/projects/beneficial-ownership-data-standard/builds/9119818/ - build failed - can you have a look at that?

Also, so this looks like it puts the translated schemas into the _build_schema or _static folder in the final website? Can we just check with @kd-ods and @ScatteredInk that's ok - there may be use cases where they want to refer people to translated schemas directly, and therefore we may care what URL they end up at

kd-ods:
Yes, to answer @odscjames. We link off to the schemas from this docs page: http://standard.openownership.org/en/v0-1/schema/schema-browser.html. (From links in text like: "View the **Ownership-or-control statement JSON schema** or explore it using the viewer below".) Ideally, we would show people the compiled, translated schema from these links, at a comprehensible URL.

docs/conf.py
rhiaro: Yes, Sphinx prints some handy output during the build so I thought this wouldn't hurt.

rhiaro:
Yep, the schema were already linked from _static, so that's where the translated schema are linked from on the schema-browser page. _build_schema shouldn't make it to the final site, that's just used during the build process by the jsonschema and csvtable directives.

kd-ods:
Noting that this is a PR into the master branch and other 0.2 work is being done on (or off) the 0.2-dev branch for eventual merging back into master. Edits to references.rst have been made on the 0.2-dev branch so conflicts will need to be resolved by hand.

odscjames:
Ideally it would be good if the changes to docs/schema/schema-browser.rst aren't in this pull request - that could cause merge conflicts later. But approved - do squash!

PR #191, opened 2019-05-24
Readme update for temporary mofile situation

rhiaro:


odscjames:
> You'll need to force add these as usually git ignores .mo files.

We could fix that in https://github.com/openownership/data-standard/blob/master/.gitignore but given that we want this to be temporary, maybe lets leave it like it is.

PR #198, opened 2019-06-11
fix a broken link

augusto-herrmann:
I came across a broken link on [this page](https://standard.openownership.org/en/v0-1/schema/guidance/index.html) when browsing the site. Let's fix it.

### DIFF
--- a/docs/schema/guidance/index.rst
+++ b/docs/schema/guidance/index.rst
@@ -3,7 +3,7 @@ Technical guidance
 
 .. attention:: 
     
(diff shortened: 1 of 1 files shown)


kd-ods:
Thanks, @augusto-herrmann. I've fixed this issue on the 0.2-dev-changelog-182 branch (so it will be fixed in the 0.2 version of the standard).

augusto-herrmann:
:ok:, so I think this PR is no longer necessary. :smile:

Issue #150, opened 2019-02-27
Updates to guidance and schema around missing information and exemptions


ScatteredInk:
Placeholder, with more detail to follow, but missing info and exemptions don't work well in 0.1 and will need to be sorted out for 0.2.

In particular: where do reasons for missing information and exemption sit?

ScatteredInk:
#149 also points to some earlier work on this that got left in to the 0.1 release.

ScatteredInk:
I have adjusted the schema to allow for information on why information is missing to sit in entity and person statements, as well as in ownership-or-control statements.

The rationale for this is that there are times when you will know what a person or entity exists but can't/aren't allowed to identify it. The metadata on why that information is missing is useful for tracking enforcement, responsibility etc.

There are also cases where a missing entity/person could appear in the middle of the structure - and in those cases we want to be able to continue building a structure despite that certainty (which is already possible) but still say why we don't the information (which isn't possible).

I've added two simple person-based examples under examples/missing-data - but the same principle would apply for entities. I think we can add to these later.

The other change I've made is to add `information-unknown-to-publisher` as a reason for missing data - the current codelist is focused on cases where we know who is _responsible_ for providing the data; this covers the case where someone is _looking_ for the data and just can't find it.

ScatteredInk:
These changes have now been made. I think there is a case for looking at different codelists at different places in the data structure for 0.3 release, so that the use of this is less ambiguous.

ScatteredInk:
Some entities are expected to have no beneficial owners, e.g. a state has legal personality but has no BOs. This should be covered by the codelist.

ScatteredInk:
Covered in #240 and #253

PR #193, opened 2019-05-29
Adds required publicationDetails component to statements - closes #135

kd-ods:
Since the new publicationDetails component is required, I've added it to all the BODS json files at the top level in the 'examples' directory. 

I've added @kindly as the main reviewer, but I think we should hold off merging this into 0.2-dev until @ScatteredInk has taken a look too.

PR #199, opened 2019-06-11
Missing information - fixes #149 and #150

### LINKED ISSUE #149: missingPersonReason and missingPersonType exist as a dependency, but are not defined
https://github.com/openownership/data-standard/blob/4f57061751e3c28319430edc7dfdaf7a38d732bf/schema/person-statement.json#L139-L142

ScatteredInk:
Discussed in more detail in #150.

@kd-ods I don't think I have followed the docs styleguide here so some attention on that would be great.

@kindly @kd-ods - your thoughts on moving missing info into both the edges and the nodes would be good. I think this messes with the purity a bit, but I don't really see a way around it atm.

### DIFF
--- a/examples/missing-data/os-01-dr-02-dc-05.json
+++ b/examples/missing-data/os-01-dr-02-dc-05.json
@@ -1,5 +1,12 @@
 [
   {
+    "publicationDetails": {
+      "publicationDate": "2019-06-13",
+      "publisher": {
+        "name": "Example Register"
+      },
+      "bodsVersion": "0.2"
+    },
     "statementID": "e9c5bedd-2627-4f16-ac0c-a1b7ebfa051d",
     "statementType": "entityStatement",
     "statementDate": "2018-12-17",
@@ -13,6 +20,13 @@
     ]
   },
   {
+    "publicationDetails": {
+      "publicationDate": "2019-06-13",
+      "publisher": {
+        "name": "Example Register"
+      },
+      "bodsVersion": "0.2"
+    },
     "statementID": "4f683208-f571-40a5-8a5a-ec78c2930387",
     "statementType": "personStatement",
     "statementDate": "2018-12-17",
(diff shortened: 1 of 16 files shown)


schema/person-statement.json
```diff
@@ -38,10 +38,36 @@
       "codelist": "personType.csv",
       "openCodelist": false
     },
-    "missingInfoReason": {
-      "title": "Missing information reason(s)",
-      "description": "For PersonStatements with the type 'anonymousPerson' or 'unknownPerson' this field should contain an explanation of the reason that detailed information on the person is not provided. This may be a standard descriptive phrase from the source system, or a free-text justification.",
-      "type": "string",
+    "unspecifiedPersonDetails": {
+      "title": "Unspecified person details",
+      "description": "When this person has a 'personType' of 'anonymousPerson' or 'unknownPerson' information explaining this should be given using the reason and description.",
```
kd-ods: This applies to the other description fields as well. Can we reformat from:

"When this person has a 'personType' of 'anonymousPerson' or 'unknownPerson' information explaining this should be given using the reason and description."

To:

"An explanation of why this person has a `personType` of 'anonymousPerson' or 'unknownPerson'. A `reason` MUST be specified."

i.e.

* back-ticks around properties
* single quotes around codes from codelists
* mentions of required child properties
* the initial part of the description being publisher/consumer-neutral: simply stating what is/should be in the field.

examples/missing-data/os-01-dr-02-dc-05.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "source": {
+      "assertedBy": [
+        {
+          "name": "Company A"
+        }
+      ]
+    }
+  }
+]
```
kd-ods: For these two new example files, we'll need to add the new publicationDetails block to them ultimately. You could add them on this branch and merge PR #193 first?

ScatteredInk:
Thanks @kd-ods.

@siwhitehouse I'm a bit unsure about this description for ooc statements:

> When confirmation has been provided that no interested party exists, where ownership and control information does not need to be provided, or where details of ownership and control are unknown, a `reason` MUST be specified. Where an unknown entity is the `subject` of further ownershipOrControlStatements in the same structure, or where there is a natural person with ownership or control but their name or details are not known or cannot be disclosed for some reason, `unspecified` should not be used, but instead a reference to a `personStatement` or `entityStatement` should be provided but identifying details MAY be left blank.

Feels like this could be shorter and maybe some of this pushed into guidance? The idea is to make sure people don't stop at ooc statement with an unspecified if an intermediary entity is unknown.

Issue #78, opened 2018-05-02
SVG text box and localization issue


kd-ods:
For localization and 'automatic' translation of SVGs we ideally would add text to diagrams and images in such a way that longer strings could be substituted in gracefully, without wrecking the image. 

In practice, this would require placing the English text string in a substantially larger text area (which didn't interfere with any other graphical elements). However, there is not yet any standard way of encoding such a text area in SVG. See ['What about flowed text?'](http://wiki.inkscape.org/wiki/index.php/Frequently_asked_questions#What_about_flowed_text) from the Inkscape documentation.

Here's the [W3C SVG 1.2 (working draft)](https://www.w3.org/TR/SVG12/).

Bjwebb:
One option to explore here would be working with whatever inkscape supports (as I've checked we can translate inkscape's flowed text). Then, for display on the website, we'd need to render as PNG. We should be able to automate this by running inkscape as part of the build process.

kd-ods:
OK - that's a reasonable solution. For the moment, since SVG with flowed text doesn't render reliably across browsers, and I've placed SVGs directly into pages, I'm not using flowed text elements.

This issue will become pressing when we start translating the docs.

Bjwebb:
Since we're not using flowed text (unless I've missed it), this isn't currently an issue for the translation.

kd-ods:
I guess the current issue for translation, @Bjwebb, is that the length of SVG text elements might not accommodate translated strings that are longer than source strings.

Bjwebb:
I think that we can mostly tweak the source SVGs so that translations fit - I've written a bit more in https://github.com/openownership/data-standard/pull/239

siwhitehouse:
I've started work on this now that we have the complete Russian translation for the diagrams. I've asked for clarification of the translation for one string as "shares owned by" has been translated as "доли в собственности / акций" which is wrecking multiple images. I'm hoping to agree a shorter alternative to this during the time I am editing the other images.

In #239 @Bjwebb said

> I've looked through all the SVGs, and I think this is true of all of the text, except for one example. In [data-schema-model-1.svg](https://github.com/openownership/data-standard/blob/237-translate-update/docs/_assets/data-schema-model-1.svg), a longer string for "Interest" really won't fit in the box. We currently have "Доля участия" which is twice as long. We could fix this by making the 2 boxes taller, but I think that might be too much whitespace on the English SVG.

Is there a way to make the box wider and have the (vertical) text wrap around on two lines?

[...]

Issue #58, opened 2018-03-08
Create a mechanism to create translatable, themable diagrams


robredpath:
- Minimum requirement is guidance for @kd-ods to start writing docs (eg recommended SVG tool or the informed belief that we can use Mermaid (or similar)) even if we need to do more work to make it pretty 

- Doing the work of prettification is out of scope, as long as it's <10days work to do such prettification

timgdavies:
Rough example of the kind of document.

![untitled drawing](https://user-images.githubusercontent.com/342624/37162118-5e009072-22ec-11e8-9bb3-1f7c7906e404.png)

Bjwebb:
I think it would be a lot of work to make mermaid.js create something that looks sufficiently like the example diagram. (Mermaid has "themes" but they're basic colour switching, I think we'd end up writing a lot of our own drawing code.) Additionally I think we'd have to translate the entire mermaid document for an object in one go.

By contrast, we can translate each piece of text in SVGs seperately (and push to transifex etc.). I've had good success using [http://itstool.org/](itstool) and [and this example its-rules.xml](https://www.w3.org/TR/SVGTiny12/i18n.html#SVGi18nl10nmarkup).

Using a [random image off wikimedia](https://commons.wikimedia.org/wiki/File:Motherboard_diagram.svg), and a pseudo translation:
```
itstool -i its-rules.xml -o Motherboard_diagram.po Motherboard_diagram.svg
podebug --rewrite unicode Motherboard_diagram.po > Motherboard_diagram_pesudo.po
msgfmt Motherboard_diagram_pesudo.po
itstool -m messages.mo -o Motherboard_diagram_translated.svg Motherboard_diagram.svg
```
![motherboard_diagram svg](https://user-images.githubusercontent.com/634/37421404-1f4aea98-27b1-11e8-9f1b-b055e61a9e3c.png)
![motherboard_diagram_translated svg](https://user-images.githubusercontent.com/634/37421412-248a55b6-27b1-11e8-814f-4a430b8cbe42.png)

This example image was created using inkscape. I've also created inkscape svgs from scratch that work well also.

Is "use inkscape" useful advice?
What's the source of the example diagram? Would it be useful to demo translating that?

kd-ods:
I think 'use inkscape' is probably realistic in the short-term. I could create a library of [objects](http://www.inkscapeforum.com/viewtopic.php?f=5&t=982), that we can re-use to create the diagrams. 

[...]

PR #239, opened 2019-08-08
Translate SVG files

Bjwebb:
#58 

Extract translations, and translate SVG files using `itstool`.

The extracted strings can be seen on Transifex at https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/svg/179539355

`itstool` is a pain to run on Read the Docs (requires libxml2), so I've added instructions to the README to generate the translated SVGs, and commit them. I've then edited `conf.py` to move the translated SVGs to the correct place in the translated build.

This doesn't deal with [flowed text](https://github.com/openownership/data-standard/issues/78#issuecomment-519080064), because we've not currently got any.

There's also some work still to be done on tweaking the SVGs - to align the text correctly, and make sure spaces for text are big enough. I think we should be able to does this by tweaking the source SVGs, in ways that won't ruin the English version.


Bjwebb:
> I think we should be able to does this by tweaking the source SVGs, in ways that won't ruin the English version.

I've looked through all the SVGs, and I think this is true of all of the text, except for one example. In [data-schema-model-1.svg](https://github.com/openownership/data-standard/blob/237-translate-update/docs/_assets/data-schema-model-1.svg), a longer string for "Interest" really won't fit in the box. We currently have "Доля участия" which is twice as long. We could fix this by making the 2 boxes taller, but I think that might be too much whitespace on the English SVG.

Bjwebb:
The `svg.mo` file doesn't need to be committed, because the translation of the SVGs happens locally.

I think I see the changes on http://standard.openownership.org/ru/58-translate-svg/schema/concepts.html, is that right?

siwhitehouse:
Thanks Ben. I'd be inclined to add a note in the instructions to say that `svg.mo` doesn't need to be added after the section on adding the `schema.mo` and `codelist.mo` files.
Yes, I see the changes too now. After I built the version I was still seeing "Aкция: 60%" in the first svg. CTRL+REFRESH didn't help either.

Issue #245, opened 2019-08-23
Remaining strings untranslated for Russian translation of docs, schema and codelists


siwhitehouse:
there are a few outstanding strings that have not been translated into Russian. These were added to Transifex after Prime Productions completed their work.

### Index
[Website](http://standard.openownership.org/ru/latest/index.html) 
[Transifex]( https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/index/179247049?q=translated%3Ano)

English | Russian
-- | --
For Newcomers |  
For Developers |  

### Schema Reference
[Website](http://standard.openownership.org/ru/latest/schema/reference.html)
[Transifex](https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/schema--reference/176065071?q=translated%3Ano)

English | Russian
-- | --
A string of minimum length 32 and maximum length 64. |  

### SVGs
The following pages have text in diagrams that is untranslated. They are all in the [same resource in Transifex](https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/svg/179539360?q=translated%3Ano)
[Website](http://standard.openownership.org/ru/latest/primer/whatisbo.html)

English | Russian
-- | --
Person A |  
100% owned by |  
Person B |  
0% shares owned by |  
100% shares owned by |  
60% directors appointed by |  
30% shares owned by |  
Person C |  
(Executive Director) |  

[Website](http://standard.openownership.org/ru/latest/primer/whyisbo.html)

English | Russian
-- | --
60% voting rights held by |  
20% shares owned by |  
50% shares owned by |  
60% directors appointed by |  

[Website](http://standard.openownership.org/ru/latest/primer/whatisthebods.html)

English | Russian
-- | --
30% shares owned by |  
99% shares owned by |  
1% shares owned by |  
standardised data |

siwhitehouse:
We will ask the second reviewer to translate these strings for us. Then we will edit them in to Transifex before going through the workflow so they appear in the repo and RTDs. The tables shown above are for illustration only.

siwhitehouse:
I suspect that I should have separated the purely textual updates from the svg files as separate branches. @odscjames could you advise, please? 

I have another commit outstanding that affects most of the svg files and svg.po. Mainly this is because Sergei spotted that we were not capitalising Компания and to correct for that. I'll wait to discuss the branching before pushing this up.

Issue #256, opened 2019-09-07
Fix markup for headings on about/index.rst


siwhitehouse:
This looks to be the same issue as https://github.com/openownership/data-standard/issues/223 that I didn't pick up at the same time.

The file contains raw blocks of html. For example

```
.. raw:: html

    <h2>
        Background
    </h2>
```

editing this so that it reads

```
.. cssclass:: h2

 Background
```

is the fix that I think we want.

siwhitehouse:
I've edited about/index.rst and the headings are [now in Transifex](https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/about--index/181814856?q=translated%3Ano) 

I've emailed Sergei, who did the review work, and asked him if he will take a look at these for us.

siwhitehouse:
*face palm*

https://github.com/openownership/data-standard/pull/258/files#diff-f3abcec1711c97466fa106ecb110a044

includes the changes to about/index.rst that I made as part of the fix for this issue.

siwhitehouse:
We've had no response from Sergei so I have used Google Translate to give us an approximation of the Russian we can use for these strings.

I haven't pushed these up as I have pushed the edited about/index.rst file in a different branch by mistake.

https://github.com/openownership/data-standard/pull/258/files#diff-f3abcec1711c97466fa106ecb110a044

PR #258, opened 2019-09-09
#245 ru-strings-svg

siwhitehouse:
This updates the svgs using the strings that will be added by #257


siwhitehouse:
> There are strings in about/index.po like 'License and contributor agreement' which have been given no RU translation string. Is that right?

No =0(

about/index.po has those because I pushed the changes to about/index.rst from #256 to this branch by mistake. This needs unpicking before this can be merged.

PR #261, opened 2019-09-19
Fix version numbers in sub-schemas

odscjames:
https://github.com/openownership/data-standard/issues/246


odscjames:
@ScatteredInk Is this what you meant, have I missed anything?

@rhiaro I don't think there is anything I need to do in terms of build processes, can you just confirm that?

PR #266, opened 2019-10-17
Update README.md

siwhitehouse:
Adding a reference to bods-dev-handbook for full set up instructions for the translation workflow


rhiaro:
The slightly nicer human readable (html rendered) link is https://openownership.github.io/bods-dev-handbook/translations.html rather than to the md file

PR #274, opened 2020-01-30
Update serialization.rst

ScatteredInk:
Fix typo in serialization docs.


kd-ods (dismissed):

Bjwebb:
@stevenday I've needed to fix the tests, because they try to download a metaschema from a url that no longer exists. I've bodged it using sed, because its the quickest thing to do for now, and the long term fix is to use the latest version of jscc.

stevenday (approved):
Thanks @Bjwebb - looks fine to me as a stopgap 🚀

Bjwebb:
Hitting merge since @kd-ods approved the text change.

Issue #277, opened 2020-02-10
Why the mixed licensing?


stevenday:
I'm looking to clarify all of the licenses on OO's projects and this one struck me as odd. I couldn't find any issues about it, so I thought I'd ask here for posterity.

I have two main questions
- Is the copyright still attributable to the World Wide Web foundation, or do we need to update that?
- Why do we use Apache for the schema and GNU Free documentation License for the documentation? Could we drop the latter and license everything under the Apache license?

Relatedly:
- Why the Apache license? I would have (naively perhaps) gone with MIT

@timgdavies - it seems like you added the licenses originally, so I'm hoping you might know the answer to these.

timgdavies:
A very good question @stevenday. I'm going to ask if @Bjwebb is able to have a quick check through Open Data Service archives for any mention of this - as I have vague recollection of discussions about the weaknesses of Apache license for documentation.

However, unless that throws up any major considerations, I would be happy with Apache 2.0 for everything, and the Copyright can be changed to belong to OpenOwnership rather than Web Foundation (who only held it on behalf of OpenOwnership).

timgdavies:
Looking more closely (and dredging up more from my memory) I think there is a substantial difference in what should be allowed in terms of:

* The schema - which users need to be free to adapt for use in their own systems;

* The documentation - where there needs to be protection against a 'rouge' re-user forking the documentation and presenting a modified version as if it is 'The Beneficial Ownership Data Standard'.

Standards are perhaps a special case in open source - as their value can be diminished by being forked.

stevenday:
Thanks for the explanation @timgdavies

So for the record, the main feature of the documentation license is that it requires people to use a different title and acknowledge the history of the document they're 'forking'. I'll close this ticket then and open a new one to fix the naming.

PR #278, opened 2020-02-11
Update copyright attribution and name in License

stevenday:
See #277 for explanation.


stevenday:
Thanks both, @Bjwebb - I've updated the year as suggested (I force pushed since it's such a tiny change and only a single commit). Would you mind re-approving when you have a moment (and feel free to merge if you like)

PR #280, opened 2020-02-17
Add Google Analytics & update privacy policy

Lathrisk:
Closes https://github.com/openownership/data-standard-sphinx-theme/issues/55


stevenday:
FYI @Lathrisk and @timgdavies - I've now removed Piwik/Matamo in this PR as well, and I've taken a first stab at updating the privacy notice regarding GA. I'll add a formal review with the next steps so it's clear why we're not merging this yet.

stevenday:
We're going to leave this open until OO have finished some ongoing work around GDPR compliance across all their sites and systems, so that we can be sure we're doing things properly.

stevenday:
I've been doing that GDPR work I mentioned (in March 😱) and I've updated this so that:

1. It's using Google Analytics setting to anonymise people's IP addresses
2. It's not allowing Google analytics to set cookies at all
3. The privacy policy page just points people to OpenOwnership's master privacy policy (which is not live yet)

In addition, I've removed hypothesis, since it's adding cookies and probably collecting data that our privacy policy doesn't over. I'll await some confirmation that this is ok before making this permanent.

On translations, this now has new privacy notice content, so will need translation into Russian and Spanish. *However*, since it's just a pointer to another page I think we can keep translating it, which simplifies things quite a bit. In the interim we can probably live with this page being in English as a single sentence with a link at the end is probably understandable enough for anyone (especially when they're going to have to read the legalese in English anyway).

I checked, and the privacy page is only on v0.2, I think we can live with just porting these changes to that version for now.

I think the outstanding things now are:
- [ ] Getting the new OO privacy policy checked and live
- [ ] Confirming if we can remove Hypothesis or if we need a cookie banner + privacy policy text to cover it

Bjwebb:
Do the Data Review Tool terms also need updating? https://github.com/openownership/cove-bods/issues/36

kd-ods:
@stevenday Hypothesis is a plugin that allows annotations to be added to text on ReadTheDocs. I think it's used as part of the review workflow for translations so maybe hold off on removing it until you've checked in with @siwhitehouse. (He's away atm.) 

As we improve the docs, I do like the idea of having an annotation mechanism for people to pose questions or request clarifications. Would any of the [info here](https://web.hypothes.is/blog/gdpr/) help us retain Hypoths.is?

stevenday:
@Bjwebb 
> Do the Data Review Tool terms also need updating? openownership/cove-bods#36

I looked at those briefly and thought that it seems standalone enough and detailed enough to leave it alone for now. The ticket doesn't make clear what exactly needs updating - it looks to be ok from my quick scan?

If you would like to transfer more legal responsibility for the service to OO, we could look at bringing it under the same policy, but I'd like to get it launched before we do that.

@kd-ods 

Thanks for the explanation and the link. At a first reading I disagree with their statements there - in my (admittedly brief) the plugin sets a cookie as soon as you load a page, whether you're logged in or not. This is not necessarily a GDPR issue, but it is a cookie law issue so they're not right to say they don't need a cookie notice. I'll investigate more if @siwhitehouse is keen to keep it, it's not the end of the world if we have to have one, just a bit of hassle to implement.

siwhitehouse:
We discussed this today and agreed that we can remove Hypothes.is.

For translation work we use the commenting function in Transifex for comments and queries. I think we have previously discussed the possibility that regular users could comment on the quality of a translation through Hypothes.is.

In practice we do not have people making comments about either the content or its translation through Hypothes.is and I suspect  this would continue to be the case unless we were to actively encourage people to use it. We don't have it installed on other data standard documentation sites that we maintain.

stevenday:
@kd-ods or @siwhitehouse - would you mind giving this a quick once over and approving? Nothing is new, but we've finally, (finally!) got the main new privacy policy up and running at: https://www.openownership.org/privacy/ so we can merge this PR too.

kd-ods (approved):
Looks good to me, Steve.

Issue #171, opened 2019-05-15
Knowledge sharing: JSON Schema and CSV Codelist tests


jpmckinney:
Since OCDS and BODS use the same mechanism for specifying the schema and codelists, I thought it'd be relevant to share scripts we've been improving since August 2017: https://github.com/open-contracting/standard-maintenance-scripts/tree/master/tests

These scripts run on our standard, profiles and extensions to ensure there are no structural errors in our schema and codelists. We use the same Travis configuration across all those repositories, while keeping the tests in one place.

The `test_*.py` files are run by pytest, and each `test_*` method is a test. Each method has a docstring, so it should be easy to understand what each test is doing. A rough summary:

JSON tests

* Ensures all JSON files are valid JSON.
* Ensures all JSON files are formatted for humans (indented).
* Ensures all JSON Schema files are valid JSON Schema Draft 4 and use codelists correctly. Ensures JSON Schema files have required metadata (e.g. title, description) and valid references. There are a variety of subtests in `validate_*` methods:
  * Prints and returns the number of errors relating to the letter case of properties and definitions.
  * Prints and returns the number of errors relating to metadata (title, description, type).
  * Prints and returns the number of errors relating to codelists in a JSON Schema.
  * Prints and returns the number of errors relating to the `type` of `items`.
  * Prints warnings relating to deep objects, which, if appropriate, should be modeled as new definitions.
  * Prints and returns the number of errors relating to objects within arrays lacking `id` fields.
* Ensures a repository has no empty files and an extension has no versioned-release-validation-schema.json file.
* Extension-specific tests

CSV tests

* Ensures all CSV files are valid: no empty rows or columns, no leading or trailing whitespace in cells, same number of cells in each row.
* Ensures all codelists files are valid against codelist-schema.json (you would need to edit the schema, as there are differences in letter-case and column headings)

Readme tests (for extensions)

* Ensures the extension's documentation contains an example.

I'd be happy to work together to factor out common tests where possible, so that we can use a common test suite for our shared way of specifying each standard. The tests can be factored out into a Python package, for example.

ScatteredInk:
Thanks for sharing @jpmckinney, this is really helpful. Once we have the current bit of schema development out of the way, we plan to have a think about testing in the standard repo - so taking this approach sounds great.

kd-ods:
Just to note that I've been getting to grips with the current testing framework in the repo as a result of [this pull request](https://github.com/openownership/data-standard/pull/193)

[...]

PR #284, opened 2020-02-25
Use flake8 and CSV tests from OCDS standard-maintenance-scripts

Bjwebb:
We already run flake8, but only against the test directly, so this includes some changes to `docs/conf.py` so that it passes.

This runs tests that all CSVs should pass. Tests specifically against codelists don't run, because the BODS codelists don't get detected as such.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,13 @@ sudo: false
 language: python
 python: 
   - "3.5"
+env:
+  global:
+    # Pinned to a commit on the master branch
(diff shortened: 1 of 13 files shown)


Bjwebb:
I agree. At the moment I just know they fail, but I'll spend some time this week breaking down the work to make them useful to us.

Bjwebb:
Looks like the list of changes to make all the tests pass is in https://github.com/openownership/data-standard/issues/171#issuecomment-523926459. I've added a link to that.

stevenday:
Looks good @Bjwebb - I think this needs a rebase because I just merged my license changes, otherwise good to go 👍

PR #289, opened 2020-03-05
Spanish translation

siwhitehouse:
This includes the translated strings from Transifex. We have some outstanding snagging queries to be dealt with separately.


siwhitehouse:
This should be merged into ```v0.2.0``` *not* ```master```

Issue #287, opened 2020-03-04
Spanish translation: SVG diagrams requiring an edit


siwhitehouse:
1. One SVG diagram has a slight missing portion of text in the Spanish translation

http://standard.openownership.org/es/latest/_images/Diag4b-splitContr2.svg

2. All of the SVGs on the [concepts page](http://standard.openownership.org/es/latest/schema/concepts.html) have had their schema propertiess translated and these will need to be reverted to the English. This can be done by editing the text in Transifex.

siwhitehouse:
2. I have changed the schema properties so they appear in English.

siwhitehouse:
1. The SVG diagram has been edited so that the Spanish version shows all of the txt. I have only pushed up the Spanish version of the file to #290

siwhitehouse:
Both issues are resolved by #290

Issue #279, opened 2020-02-13
Create a process for ad hoc updates in between schema versions


siwhitehouse:
We do not currently have an agreed mechanism for making ad hoc updates to the website in between schema versions. We need this so that we maintain consistency between the English and translated versions of the website.

stevenday:
@siwhitehouse - my assumption is that as soon as we change any text, this will have to follow the same process as schema releases (in terms of freezing, translating, releasing). Obviously with less need for review, example testing, etc since we're not changing the schema/standard itself. 

Elaborating on the options gives four kinds of release:

- No text changes: no translation needed, no new version, release immediately
- Text changes but no schema changes: translation needed, freeze and translate release process
- Minor schema changes (bug fixes, typos, etc): translation, new patch version, freeze and translate release process
- Schema changes: translation, new major or minor version depending on changes, full review and release process

Does that sound accurate?

Aside: I think some better tagging of released versions, and perhaps using a [release branch](https://nvie.com/posts/a-successful-git-branching-model/) system would help if we need to separate these kinds of changes in the future, and avoid the need to 'freeze' development on master.

stevenday:
Talking in our sprint planning it seems like there's some appetite for a better git branching system to help manage translations and release fixes etc. @Bjwebb - do you have thoughts on this and whether it's something we could prioritise in our sprint call tomorroqw?

I also note there's a smaller related ticket in https://github.com/openownership/bods-dev-handbook/issues/13 about branch names, which ties into another question I was going to ask about tags.

I note that we're very explicit about using branches, not tags: https://github.com/openownership/bods-dev-handbook/blob/master/standard_releases.md#versioned-releases but I'm not sure why that is? I was surprised when I tried to find the canonical v0.2 in github that it was a branch, not a tag.

odscjames:
> I note that we're very explicit about using branches, not tags: https://github.com/openownership/bods-dev-handbook/blob/master/standard_releases.md#versioned-releases but I'm not sure why that is? I was surprised when I tried to find the canonical v0.2 in github that it was a branch, not a tag.

This was to allow the possibilities of technical updates that did **NOT** change content being made to each version. 

For instance, you can see we have used this option to add a banner redirecting people to the latest version from older versions (I looked into using the RTD features for this, and it turns out we couldn't because we had named our tags "wrong", and in the end I determined it was less work and less disruptive to just add a banner by hand to each branch.)

siwhitehouse:
[...]

PR #290, opened 2020-03-05
Spanish translation

siwhitehouse:
This includes the translated strings from Transifex. We have some outstanding snagging queries to be dealt with in separate issues.


siwhitehouse:
@kd-ods It's dawned on me that we may need to decide how to update existing versions of the schema before we merge this. 

Here I am merging to ```v0.2.0```  but I'm not clear how this branch would then be brought in to ```master```.

One option, that you spoke about yesterday, is that we merge into ```master``` and then cherry-pick commits so that they are applied to the version (```v0.2.0``` in this case).

Connected to the discussion we are having on #279

siwhitehouse:
@kd-ods I've changed this so it merges into ```master``` as per discussion with @stevenday on #279

siwhitehouse:
I've pushed up the changes that I have made to the SVG documents as per #287 @kd-ods could you review now please?

PR #291, opened 2020-03-06
translation setup: use ocds-babel instead of bods-babel

Bjwebb:
https://github.com/openownership/bods-babel/issues/4

I've checked that .pot files and translated html don't have any changes besides timestamps and version numbers.

Once this is merged we can archive the bods-babel repo.


stevenday:
> Do we also need to update requirements.in though @Bjwebb? It also points at bods-babel...

Sorry, I was looking at the wrong branch here, all good!

Bjwebb:
I've rebased onto master.

PR #295, opened 2020-03-11
Fixing the translation of titles and codes in the schema #294

### LINKED ISSUE #294: Spanish translation: Schema reference titles and codes
All elements of the schema have been translated. We need to edit this so that titles have the format 

EnglishVersion (SpanishVersion)
eg
AddressType (TipoDeDomicilio)

Codes need to be retained in their original English.

siwhitehouse:
This fixes #294


siwhitehouse:
Thanks @kd-ods

Issue #285, opened 2020-03-04
Spanish translation: sphinx-theme updates not showing


siwhitehouse:
See the [navigation bar](http://standard.openownership.org/es/latest/) where the placeholder English text is still visiible.

I think this could be because I have built the ```spanish-translation``` branch on the Beneficial Ownership Data Standard (ES) project, and not on the main Beneficial Ownership Data Standard project. 

I'm currently blocked on testing this as [the builds are failing](https://readthedocs.org/projects/beneficial-ownership-data-standard/builds/)

siwhitehouse:
Builds are still failing. @Bjwebb will look at this and then I will build the main project.

Bjwebb:
I'm not sure what's going on here. Its failing early in the build process (whilst still fetching stuff from git, so before actually running our code!), and there's not much error information. The same commit seems to have built fine a few weeks ago.

I've reported the issue on the readthedocs GitHub: https://github.com/readthedocs/readthedocs.org/issues/6754

siwhitehouse:
The builds are now working, but the sphinx-theme updates are not showing. This is because we need to update requirements.txt with the latest commit in the data-standard-sphinx-theme repo.

siwhitehouse:
Fixed by #301 
Following the merge I updated the RTDs build and the navigation bar menu is now showing the translated strings.
We will need to update the dev-handbook wrt ad hoc updates to document that any changes to the navigation bar in sphinx-theme will also need the branch containing the latest version needs to reference the latest commit in requirements.txt

PR #301, opened 2020-03-12
Update requirements.txt

siwhitehouse:
Backdating the sphinx theme commit in requirements.txt so that the navigation bar translations are pulled through to RTDs.


siwhitehouse:
This fixes #285

PR #303, opened 2020-03-12
Use some of the json schema tests from ocds

Bjwebb:
These tests are being moved to the jscc module, from standard-maintenance-tests: https://github.com/open-contracting/jscc/pull/1

This PR creates a `test_json.py` file which imports some of those tests, and does some BODS customisations. To make those tests pass, it also indents the schema files.~~, and removes some `propertyOrder` keys ([see issue](https://github.com/openownership/data-standard/issues/171#issuecomment-598195791)), but we can keep those if we want~~


Bjwebb:
Currently this uses an unmerged jscc commit, so we might want to wait for that to be merged before merging this.

jpmckinney:
FYI, I filled in jscc, now documented at https://jscc.readthedocs.io/en/latest/ See in particular: https://jscc.readthedocs.io/en/latest/api/testing/checks.html

Bjwebb:
> Just for my understanding, where/how do we get jscc into the python virtualenv? e.g. to run these tests locally. I don't see it in any of the requirements files?

As discussed on the call this morning, the pip install happens [inside the install.sh](https://github.com/openownership/data-standard/blob/master/.travis.yml#L11) which is [pulled in from standard-maintenance-scripts](https://raw.githubusercontent.com/open-contracting/standard-maintenance-scripts/c3b4fc9499852baed9fc0160638162aa0e61381f/tests/install.sh). I've [added a note](https://github.com/openownership/data-standard/issues/306#issuecomment-600057952) about dropping this once we no longer use the standard-maintenace-scripts repo for the csv tests.

ScatteredInk:
@kindly I think this is the change that has resulted in a bit of fiddling around to get tests running locally.

jpmckinney:
FYI, as linked above and suggested by @Bjwebb, it might be less fiddly to switch to https://jscc.readthedocs.io/en/latest/ instead of using a specific commit of standard-maintenance-scripts.

jpmckinney:
The most relevant docs page is: https://jscc.readthedocs.io/en/latest/api/testing/checks.html

PR #302, opened 2020-03-12
Theme rebrand (onto master)

stevenday:
These are the changes necessary to bring https://github.com/openownership/data-standard-sphinx-theme/pull/57 into the master branch of the standard.

@Lathrisk did all the work, I've just squashed the WIP commits with various updates to the theme hash into one, with the most recent hash. I'll update this again with the commit hash of where that PR is merged into master, so this has to wait for that to be merged first.

We had to change some colors in the SVGs in this commit, I've made it look like @Lathrisk did the new Spanish ones too, although they actually came into existence in parallel. 

Note: I assume this is the desired process @Bjwebb / @Lathrisk ? Let me know if not.

I've set up similar PRs to merge this change into the [v0-2-0](#304) and [v0-1](#305) release branches.

PR #304, opened 2020-03-12
Theme rebrand (onto v0-2-0)

stevenday:
Counterpart to #302, to put the new theme into v0-2-0.


stevenday:
I've updated this now that I've merged the theme changes into master. @Bjwebb would you mind giving this a quick review/approval?

stevenday:
> This is just the Spanish SVGs, is that correct? 

Hmm, no, those should be in there but in addition to the existing ones. Let me fix that.

stevenday:
@Bjwebb - e342951 now includes the original SVGs and the Spanish & Russian ones. Thanks for catching that!

PR #305, opened 2020-03-12
Theme rebrand (onto v0-1)

stevenday:
Counterpart to #302, to put the new theme into v0-1.

Here I haven't been able to directly cherry-pick @Lathrisk's work, so I've made commits approximating them:
- Updating the theme commit hash (which is pinned to a specific commit in both `requirements.in` and `requirements.txt` in `v0-1`)
- Updating all the SVG diagrams (which are older versions pre-translation work and we don't have language specific built versions)

I've also made a small change to `docs/conf.py` (The ReadTheDocs config file) to backport https://github.com/openownership/data-standard/commit/980ec78787e50dfec756e0d91931faa371c1b47e (adding a variable for the version number so we can display it in the sidebar).

Note that this currently removes the "old version" banner added by https://github.com/openownership/data-standard-sphinx-theme/commit/d5bbf2b4f2e67866918ed8194b57b5f7bebbafae. I don't think keeping separate theme branches for each old version is very sustainable (and I note we don't do it for the three v0-0-x branches already) so I'm going to explore tackling https://github.com/openownership/bods-dev-handbook/issues/13 in the first instance to replace this.

### DIFF
--- a/docs/_assets/Diag1-simpleOwn.svg
+++ b/docs/_assets/Diag1-simpleOwn.svg
@@ -76,7 +76,7 @@
       <path
          id="path4371"
          d="M 0,0 5,-5 -12.5,0 5,5 0,0 Z"
-         style="fill:#00deb6;fill-opacity:1;fill-rule:evenodd;stroke:#00deb6;stroke-width:1pt;stroke-opacity:1"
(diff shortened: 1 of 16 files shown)


stevenday:
I've updated this now that I've merged the theme changes into master. @Bjwebb would you mind giving this a quick review/approval?

PR #310, opened 2020-03-24
Add a pull-request template

stevenday:
To signpost the dev handbook better and make sure we think about
translations and releasing at merge time.

In the various changes we've made to process and docs recently this
came up a few times as a way to improve awareness and it's easy to add.


rhiaro (dismissed):
This seems like a good idea. I wonder if it's worth mentioning making sure any relevant documentation and/or changelogs are updated if required as well?

kd-ods:
Good point. I've added checklist items. Please re-review :-)

(I think I might put sthg in the BODS handbook about maintaining the changelog and link to https://keepachangelog.com/en/0.3.0/.)

rhiaro (approved):

PR #325, opened 2020-07-27
Delete flat-template.xlsx

kd-ods:
This flat template is out-of-date. Deleting as per internal issue https://www.notion.so/openownership/Update-BODS-flat-template-for-internal-use-and-remove-from-Standard-repo-90b5521bd4ff4922811cce71d7bdb639.


stevenday:
For future reference - in the linked internal issue we discussed the use of this flat template and decided it wasn't something we wanted to support as part of the standard at this point (it's not polished enough, and not necessarily a workflow we'd recommend anyway). Given that it was also out of date, it was therefore better to remove it from the repo.

PR #330, opened 2020-10-09
Text improvements in Primer

kd-ods:
Prompted by the glossary work. We can make some minor edits which mean certain terms can be removed from the glossary

### DIFF
--- a/docs/primer/whatdoesbodscover.rst
+++ b/docs/primer/whatdoesbodscover.rst
@@ -43,4 +43,4 @@ In order to enable worldwide use of the Standard, it does not directly define th
 
 .. highlights:: 
     
-    The Standard is well-defined yet flexible. National registers of beneficial ownership can use it; so can people researching or representing corporate structures.
+    The Standard is well-defined yet flexible. National registers of beneficial ownership can use it; so can people researching or describing corporate structures.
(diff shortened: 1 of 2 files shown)


docs/primer/whatdoesbodscover.rst
```diff
@@ -43,4 +43,4 @@ In order to enable worldwide use of the Standard, it does not directly define th
 
 .. highlights:: 
     
-    The Standard is well-defined yet flexible: as suited for use by national registers of beneficial ownership as by corporate researchers.
+    The Standard is well-defined yet flexible. National registers of beneficial ownership can use it; so can people researching or representing corporate structures.
```
siwhitehouse, line 46: "representing corporate structures" could mean describing them or acting on behalf of them. I know what you mean, and I think the reader can work it out, but I think this might read better as "so can people researching corporate structures or those who need to describe them using BODS" or similar

siwhitehouse (changes requested):

docs/primer/whatdoesbodscover.rst
kd-ods: Have replaced 'representing' with 'describing'.

siwhitehouse (approved):
Thank you. That looks clearer.

PR #331, opened 2020-11-03
Edits to clarify sections of the website 

siwhitehouse:
and also to enable the removal of redundant terms from the glossary


docs/schema/guidance/statement-identifiers.rst
```diff
@@ -3,7 +3,7 @@
 Statement Identifiers
 =====================
 
-Each statement MUST have an entirely unique identifier. There MUST be no possibility of collision between identifiers for different statements. 
+Each statement MUST have an entirely unique identifier. Statement identifiers MUST not change over time.
```
kd-ods, line 7: I think just remove this sentence all together: "Statement identifiers MUST not change over time."

Instead, edit the first, subsequent bullet:

"Two **different** statements SHOULD never have the same identifier (concurrently OR at different points in time);"

kd-ods (commented):
One suggested amendment.

PR #332, opened 2020-11-12
Backport Google Analytics and updated privacy page to v0.2.0

stevenday:
This is identical to #280 but back-ported to the 0.2.0 branch. This is the only version branch which has a privacy page and hence needs this back-porting, I don't think it's worth trying to add either the privacy page or google analytics to older versions.


stevenday:
Sorry @kd-ods - would you mind double checking this too, if you have time?

kd-ods (approved):

Issue #326, opened 2020-09-17
Publishing information about market listings for PLCs


kd-ods:
BO registers need to publish a minimum set of information about each publicly listed company (PLC) in scope for declaration or disclosure. That includes standard company info plus the stock markets on which it is listed and the financial instruments listed there. This is how I suggest that that information is published in BODS.

Under entityStatement:

```
"publicListing": {
    "companyFilingsURLs": ["https://www.londonstockexchange.com/stock/AZN/astrazeneca-plc/analysis", "https://www.astrazeneca.com/investor-relations/stock-exchange-announcements.html"],
    "securitiesListings": [{
        "stockExchangeName": "London Stock Exchange",
        "stockExchangeJurisdiction": "GB",
        "marketIdentifierCode": "XLON",
        "securitiesListed": [{
            "ticker": "AZN",
            "isin": "GB0009895292"
        }]
    },
    {
        "stockExchangeName": "London Stock Exchange - MTF",
        "stockExchangeJurisdiction": "GB",
        "marketIdentifierCode": "XLOM",
        "securitiesListed": [{
            "ticker": "0A4J",
            "isin": "US0463531089"
        }]
    }]
}
```

Notes
-----

1. 'companyFilingsURLs' - URL or URLs where regulatory filings related to major holdings can be retrieved. URLs may point to pages maintained by regulatory bodies, stock exchanges or by the company itself.

2. 'securitiesListings' - A complete list of equity securities and the exchanges on which they are traded.

3. 'stockExchangeName' - The stock exchange or stock exchange segment where the company has listed securities.

4. I was tempted to include a url property within each securities listing (i.e. for each stock exchange). But (a) I'm not sure how useful it is to maintain the association between a filings url and the exchange, rather than putting it in the more general 'companyFilingsURLs' array and (b) with a segmented exchange like LSE there's only one page for a company, even if it has securities listed on different segments, so the url might be repeated.

5. There's also the issue of publishing some kind of flag on the entityStatement to indicate whether the entity has been exempted from standard BO disclosure on the basis that they are listed on a well-regulated exchange. I've not covered that here, since I doubt we want to plan it in isolation. There may be a whole set of 'entity features' under which 'elible-for-bo-disclosure-exemption' would be just one (boolean) property.

6. The ultimate question is probably: in general, is the scope and structure of the data above adequate to reach relevant information about a company's major holdings (with an automated process if possible)?

Hearing from those who are working with PLC data would be really useful.

[...]

PR #346, opened 2021-03-24
Schema - Add PLC information.

kindly:
# Overview
Add PLC information.

https://github.com/openownership/data-standard/issues/326

## Translations

## Documentation & Release

- [ ] I've added or edited any related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md

### DIFF
--- a/docs/schema/changelog.rst
+++ b/docs/schema/changelog.rst
@@ -15,6 +15,10 @@ ChangeLog
 [Unreleased] - 2021-03-30
 =========================
 
+Added
+-----
+- Support for describing the traded securities and status of a publicly listed company (PLC): a new ``publicListing`` object has been added to Entity Statements.
+
 Changed
 -------
 - The ``interestType`` and ``unspecifiedReason`` codelist codes have been changed from using hyphens to camelCase.
--- a/schema/components.json
+++ b/schema/components.json
@@ -565,6 +565,106 @@
           ]
         }
       ]
+    },
+    "PublicListing": {
+      "type": "object",
+      "title": "Public listing",
+      "description": "Details of a publicly listed company, its securities (shares and other tradable financial instruments related to the entity), and related regulatory filings.",
+      "required": [
+        "securitiesListings",
+        "hasPublicListing"
+      ],
+      "minProperties": 1,
+      "properties": {
+        "hasPublicListing": {
+          "type": "boolean",
+          "title": "Has public listing",
+          "description": "This value MUST be true if the entity is known to be a publicly listed company."
+        },
+        "companyFilingsURLs": {
+          "type": "array",
+          "title": "Company filings URLs",
+          "description": "URL or URLs where regulatory filings related to major holdings can be retrieved. URLs may point to pages maintained by regulatory bodies, stock exchanges or by the company itself.",
+          "items": {
+            "type": "string"
+          },
+          "format": "uri"
+        },
+        "securitiesListings": {
+          "type": "array",
+          "title": "Securities listings",
+          "description": "Details of the entity’s securities and the public exchanges and markets on which they are traded. All equity securities SHOULD BE listed here, plus any other securities from which beneficial ownership might be derived. Where a security is traded on more than one market, there SHOULD BE an entry for each market (or market segment).",
+          "items": {
+            "title": "Securities listing",
+            "description": "Details of a security and the market on which it is traded.",
+            "$ref": "#/definitions/SecuritiesListing"
+          }
+        }
+      }
+    },
+    "SecuritiesListing": {
+      "type": "object",
+      "title": "Securities listing",
+      "description": "Details of a security and the market on which it is traded.",
+      "required": [
+        "stockExchangeJurisdiction",
+        "security",
+        "stockExchangeName"
+      ],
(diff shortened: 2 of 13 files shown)


schema/components.json
```diff
@@ -565,6 +565,105 @@
...
+          "type": "boolean",
+          "title": "Has public listing",
+          "description": "This value MUST be true if the entity is known to be a publicly listed company."
+        },
+        "companyFilingsURLs": {
+          "type": "array",
+          "title": "Company filings URLs",
+          "description": "URL or URLs where regulatory filings related to major holdings can be retrieved. URLs may point to pages maintained by regulatory bodies, stock exchanges or by the company itself.",
+          "items": {
+            "type": "string"
```
kd-ods, line 589: Should we add '"format": "uri"' here too?

kd-ods:
Also, @kindly, can you update the changelog on this branch? Cheers.

Let's ask @siwhitehouse to review this PR too, since I'm a bit close to the work.

schema/components.json
kindly: yep added

schema/components.json
```diff
@@ -565,6 +565,106 @@
...
+            "description": "Details of a security and the exchange, trading platform or market segment on which it is traded.",
+            "$ref": "#/definitions/SecuritiesListing"
+          }
+        }
+      }
+    },
+    "SecuritiesListing": {
+      "type": "object",
+      "title": "Securities listing",
+      "description": "Details of a security and the exchange, trading platform or market segment on which it is traded.",
```
siwhitehouse, line 608: I think this is a little vague. From the schema I can see that  "stockExchangeJurisdiction", "security" and "stockExchangeName" are required and so they MUST be included. I'm not sure from this description when the trading platform and/or market segment SHOULD be included.

siwhitehouse (commented):
There's only the one comment I have. Could we improve the description of the "SecuritiesListing" please/

schema/components.json
kd-ods: To address this we could add to the two child MIC field descriptions the sentence:

"MIC values SHOULD be published where they are known by the publisher."

But I wonder whether this goes without saying for all non-required fields, @siwhitehouse?

siwhitehouse: It's really just in the context of this description that I think we could be clearer. I was using MUST and SHOULD to explain my concern, rather than suggest we use them explicitly  in the description. Sorry if that wasn't clear. I was thinking that something like

"Details of a security and the exchange, which may include the trading platform and/or market segment on which it is traded."

might work.

kd-ods: @siwhitehouse - I'm not sure that works. I wonder if the description of stockExchangeName, as it stands, helps.

"The name of the exchange, market or trading platform on which the security is traded. If the security is traded on a segment of the exchange, then the name SHOULD include both elements. For example, 'London Stock Exchange - MTF'."

My understanding is that stock exchanges are now considered just one type of regulated trading platform. We're using the term 'stockExchangeName' since it's more meaningful to people, but the description makes clear that it should be the name of whatever the regulated trading platform is (even if it's not a traditional 'stock exchange'). The MIC fields give additional useful identifiers for the 'exchange, market or trading platform on which the security is traded'.

siwhitehouse: @kd-ods My apologies, I don't think I'm being very clear and it feels like you are analysing this more than I am. 

I am reading

> Details of a security and the exchange, trading platform or market segment on which it is traded.

to mean that **one, and only one** from the exchange, the trading platform or the market segment are included. That's all I am suggesting that we try and resolve.

kd-ods: Ahhhhhh. Understood. Would simplifying the description to the following work, then:

"Details of a security and the market on which it is traded."

kd-ods: @kindly - that string has now been updated in the schema-generating sheet. It's ready to pull in.

kindly (commented):
@kd-ods  could you add codelist descriptions to  https://docs.google.com/spreadsheets/d/1QqzqvFJYH04PZxE4k_HCh6vd_GG1mXE5DY-FHRN7CR8/edit#gid=1183290269

The new csv tests fail if there is not a description.

kd-ods:
Descriptions now added, @kindly. (Glad our test of the tests worked out ;-) )

docs/schema/changelog.rst
```diff
@@ -15,6 +15,10 @@ ChangeLog
 [Unreleased] - 2021-03-30
 =========================
 
+Added
+-----
+- Support for describing the Public listing of an entity which is a Publicly Listed Company (PLC), using a new ``publicListing`` object.
```
kd-ods, line 20: Could you edit this to:

"Support for describing the traded securities and status of a publicly listed company (PLC): a new ``publicListing`` object has been added to Entity Statements."

Thanks, @kindly.

kd-ods (changes requested):
Just one little request on the changeLog wording.

kindly:
@kd-ods Change has been made.

kd-ods (approved):
Looks good. Thanks, @kindly. Let's go ahead and merge.

Issue #345, opened 2021-03-23
Move hasPEPStatus inside the pepStatusDetails block


ScatteredInk:
This will require validation so that the object can't be empty.

This change is so that we have the same pattern as for listed companies.

kd-ods:
`hasPepStatus` block is currently:

```
    "hasPepStatus": {
      "title": "Politically Exposed Person Status",
      "description": "Is the person described in this statement a politically exposed person? This field should not be used if PEP declarations are not expected as part of this disclosure. If a PEP declaration is expected but missing this field should not be used but the reason for the missing data declared in the pepStatusDetails field.",
      "type": "boolean",
      "propertyOrder": 79
    }
```

It sits alongside `pepStatusDetails`, which is currently an array:

```
    "pepStatusDetails": {
      "title": "Politically Exposed Person Status Details",
      "description": "One or more descriptions of this person's Politically-Exposed Person (PEP) status.",
      "type": "array",
      "items": {
        "$ref": "components.json#/definitions/PepStatusDetails"
      },
      "propertyOrder": 80
    }
```

I suggest:

- rename`pepStatusDetails` to `pepStatus` and make it an object.
- add a new `details` property at the top level of `pepStatus`, making this what is currently the `pepStatusDetails` array (title, description and all)
- move `hasPepStatus` into the `pepStatus` block at the top level, alongside the `details` block.
- make `hasPepStatus` a required field
- edit the `hasPepStatus` description to be "`true` or `false` according to whether the person described by this statement has the status of politically exposed person. A `null` value means a PEP status declaration is expected but missing; the reason for the missing data SHOULD be supplied in the `details` field." 
- title and description for `pepStatus`:

title: Politically Exposed Person status
description: Information about whether, and how, the person described by this statement is politically exposed. This field should not be used if PEP declarations are not expected as part of beneficial ownership declarations.

After this is all done, the relevant test data will need to be updated too.

kd-ods:
Closed by PR #348

PR #348, opened 2021-03-25
345 move pepstatus

kindly:
# Overview
Schema:  Move hasPepStatus into object.

https://github.com/openownership/data-standard/issues/345

## Translations

## Documentation & Release

- [ ] I've added or edited any related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md

### DIFF
--- a/docs/schema/changelog.rst
+++ b/docs/schema/changelog.rst
@@ -18,6 +18,7 @@ ChangeLog
 Changed
 -------
 - The ``interestType`` and ``unspecifiedReason`` codelist codes have been changed from using hyphens to camelCase.
+- ``hasPepStatus`` and ``pepDetails`` are replaced with ``politicalExposure`` object  that contains ``status`` and ``details`` properties.
 
 
 [0.2] - 2019-06-30
--- a/schema/person-statement.json
+++ b/schema/person-statement.json
@@ -144,21 +144,6 @@
       },
       "propertyOrder": 60
     },
-    "hasPepStatus": {
-      "title": "Politically Exposed Person Status",
-      "description": "Is the person described in this statement a politically exposed person? This field should not be used if PEP declarations are not expected as part of this disclosure. If a PEP declaration is expected but missing this field should not be used but the reason for the missing data declared in the pepStatusDetails field.",
-      "type": "boolean",
-      "propertyOrder": 79
-    },
-    "pepStatusDetails": {
-      "title": "Politically Exposed Person Status Details",
-      "description": "One or more descriptions of this person's Politically-Exposed Person (PEP) status.",
-      "type": "array",
-      "items": {
-        "$ref": "components.json#/definitions/PepStatusDetails"
-      },
-      "propertyOrder": 80
-    },
     "publicationDetails": {
       "title": "Publication details",
       "description": "Information concerning the original publication of this statement.",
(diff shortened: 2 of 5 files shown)


kd-ods:
@ScatteredInk - FYI. If you look at the example JSON files in this PR you'll see how we've brought hasPEPstatus into line with how how we're dealing with PLCs. Not assigning you explicitly as a reviewer since i know you've got a lot on.

schema/person-statement.json
```diff
@@ -182,6 +167,32 @@
...
+    },
+    "pepStatus": {
+      "type": "object",
+      "title": "Politically Exposed Person status",
+      "description": "Information about whether, and how, the person described by this statement is politically exposed. Use this property only if PEP declarations are expected as part of beneficial ownership declarations.",
+      "required": [
+        "hasPepStatus"
+      ],
+      "properties": {
+        "hasPepStatus": {
```
kd-ods, line 179: @kindly - do go ahead and adjust this to a string/enum as you suggested in conversation, so that we have an 'unknown' value. I think we prob have to massage the naming too, so that we don't end up with a pseudo-boolean field with confusing "true"/"false" string values. How about 'personsPepStatus' with values 'isPep', 'isNotPep' and 'unknown'. I'll suggest a rewording of the description below.

schema/person-statement.json
```diff
@@ -182,6 +167,32 @@
...
+          "type": [
+            "boolean",
+            "null"
+          ],
+          "title": "Politically Exposed Person status",
+          "description": "This value is 'true' or 'false' according to whether the person described by this statement has the status of politically exposed person. A null value means a PEP status declaration is expected but missing; the reason for the missing data SHOULD be supplied in the details field."
```
kd-ods, line 185: If the suggested wording above works, then this description could be:

"This value is 'isPep' or 'isNotPep' according to whether the person described by this statement has the status of politically exposed person. An 'unknown' value means a PEP status declaration is expected but missing; the reason for the missing data SHOULD be supplied in the details field."

schema/person-statement.json
kd-ods: After a chat with @kindly the plan is to rename some properties to make this align more neatly.

schema/person-statement.json
kd-ods: Agreed with @kindly that this description works.

docs/schema/changelog.rst
```diff
@@ -18,6 +18,7 @@ ChangeLog
 Changed
 -------
 - The ``interestType`` and ``unspecifiedReason`` codelist codes have been changed from using hyphens to camelCase.
+- ``hasPepStatus`` and ``pepDetails`` are replaced with ``politicalExposure` object  that contains `status` and `details` properties.
```
kd-ods, line 21: Backquotes need corrected to:

``hasPepStatus`` and ``pepDetails`` are replaced with ``politicalExposure`` object  that contains ``status`` and ``details`` properties.

kd-ods (commented):
See backquotes comment. 

Also - sorry - just spotted that we need to edit this description in components.json: https://github.com/openownership/data-standard/blob/7894f63b977466ac3212f0517a0cf1b4329ce8aa/schema/components.json#L477

It now needs to read:

"description": "An explanation of the reason that PEP status for the person is not provided (i.e. `politicalExposure.status` is 'unknown'). This may be a standard descriptive phrase from the source system, or a free-text justification. Where this field is present it should be the only field except for `source`."

(I would have made these changes myself, but figure you're doing something clever to squash commits, so I'd leave you to do it.)

kd-ods (approved):
Thanks, @kindly. Looks good!

Issue #341, opened 2021-03-01
Fix inconsistent formatting in codelists


ScatteredInk:
Most of our codelists use camelCase.

Two of our codelists, unspecifiedReason and interestType use a dash delimited format.

I think it makes sense to have everything use camelCase and update the style guide, where this issue was first raised.

kd-ods:
@rhiaro - assigning to you to do asap since the documenting aspect will unblock a lot of other work this sprint :-)

There are some related meta-issues, which I've also assigned to you, about documenting the process for updating and managing codelists, and the styleguide:

- https://github.com/openownership/bods-dev-handbook/issues/6
- https://github.com/openownership/bods-dev-handbook/issues/21

To get you started, take a look at the [codelists placeholder info here](https://github.com/openownership/bods-dev-handbook/blob/master/compiling_schema.md#codelists).

PR #349, opened 2021-03-26
Fix codelist capitalisation and csv/enum sync test

rhiaro:
Updates the codelists which used hyphens to use camelCase instead, for #341.

Also adds tests to check:

* All codelists CSVs are present somewhere in the schema, and links in schemas to codelist CSVs actually exist.
* The schema enums for closed codelists match the codes in the CSV files.
  * except for `statementType.csv` as the schemas which point to this don't use all of the values in their enums by design.

See the test run of the second commit for what they look like failing (if they're out of sync).

And I noticed that `person-statement.json` points to the `statementType.csv` codelist, but `entity-statement` and `ownership-or-control-statement` did not, so I updated the latter two for consistency.


kd-ods (commented):
I've glanced through the CSV and test data changes and they look good. I'll leave @kindly to check the tests and approve for merging. I think the changeLog just needs to note that the format/style of codelist values has been made consistent.

rhiaro:
@kd-ods Changelog updated

kindly (approved):

PR #358, opened 2021-07-30
Remove BODS roadmap image, separate OpenOwnership name

StephenAbbott:
# Overview

Remove BODS roadmap image, separate OpenOwnership name

@kd-ods 

(Closes/Relates to) issue: #

## Translations

## Documentation & Release

- [x] I've added or edited any related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md

### DIFF
--- a/docs/about/index.rst
+++ b/docs/about/index.rst
@@ -13,15 +13,12 @@ Money laundering, tax evasion and obscure chains of ownership are transnational 
 
 The Beneficial Ownership Data Standard enables this transparency by providing a solid conceptual and practical framework for collecting and publishing beneficial ownership data, resulting in data that is interoperable, more easily reused, and of higher quality.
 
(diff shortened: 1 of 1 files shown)


kd-ods:
I worry here that the changes in the English version mean that strings won't have a match in the .po files, and hence will break the translations. I think it's worth having Simon review this on his return since he's got a better grasp on how the translation files work.

siwhitehouse (approved):
I'm inclined to approve this. My experience with Transifex is that these changes are enough to break the link between the original text and the translation. However, we do have access to the original translations in Transifex and the changes are obvious enough for us to have some confidence that we can replace them.

We are going to have to do that work at some point in the future and it seems worthwhile to do it now - and get Open Ownership's name right.

I've checked this branch locally and the removal of the image hasn't broken any of the text.

Issue #360, opened 2021-09-23
Feature: Represent State-Owned Enterprises (SOEs)


kd-ods:
_[This ticket helps track progress towards developing a particular feature in the Beneficial Ownership Data Standard (BODS) where changes or revisions to the standard may be required. It should be placed on the [BODS Feature Tracker](https://github.com/openownership/data-standard/projects/4), under the relevant status column._

_See [Feature development in BODS](https://openownership.github.io/bods-dev-handbook/feature_development.html) in the Handbook._

_The information in this first post on the thread should be updated as necessary so that it holds up-to-date information. Comments on this ticket can be used to help track high-level work towards this feature or to refine this set of information.]_

### Feature background

**What user needs are met by introducing or developing this feature in BODS?** _[Summarise these needs. Link to user stories, reports, blogs and other evidence where possible. Or add user stories here directly.]_

People publishing and interpreting BODS data should be able to:

- Model SOEs in a way that allows for accurate identification and useful analysis. This should include modelling the jurisdiction and, where appropriate, the state body that controls the entity.
- Identify a declaring or an intermediary entity as an SOE, either directly (within a declaration) or by aggregating declarations.
- Declare that an entity is exempt from declaration because it is an SOE, while retaining as much information about the controlling jurisdiction as is available.
- Potentially, describe control within SOEs with appropriate interest types.

The [Open Ownership Principles](https://www.openownership.org/principles/comprehensive-coverage/) suggest that disclosure regimes include state-owned-enterprises (see [related blogpost from January 2021](https://www.openownership.org/blogs/state-owned-enterprises-a-new-frontier/)). And Open Ownership has published [recommendations](https://www.openownership.org/blogs/state-owned-enterprises-and-beneficial-ownership-disclosures) around collecting and publishing information about state-owned enterprises. As a minimum BODS should enable publishers to implement those recommendations.

In 2021, Open Ownership launched the [Opening Extractives programme](https://www.openownership.org/news/collective-action-on-extractive-sector-corruption-can-turn-the-tide/), a five-year joint partnership with the Extractives Industries Transparency Initiative (EITI). [Requirement 2.6 of the EITI Standard](https://eiti.org/document/eiti-standard-2019#r2-6) outlines disclosure of ownership requirements related to SOEs operating in the oil, gas and mining sector. Better modelling of the ownership of SOEs is crucial so that members of national EITI multi-stakeholder groups can judge whether or not countries and extractives companies are fulfilling this requirement of the EITI Standard.

[...]

Issue #359, opened 2021-09-23
Feature: Represent market listing information for companies (PLCs)


kd-ods:
_[This ticket helps track progress towards developing a particular feature in the Beneficial Ownership Data Standard (BODS) where changes or revisions to the standard may be required. It should be placed on the [BODS Feature Tracker](https://github.com/openownership/data-standard/projects/4), under the relevant status column._

_See [Feature development in BODS](https://openownership.github.io/bods-dev-handbook/feature_development.html) in the Handbook._

_The information in this first post on the thread should be updated as necessary so that it holds up-to-date information. Comments on this ticket can be used to help track high-level work towards this feature or to refine this set of information.]_

### Feature background

**What user needs are met by introducing or developing this feature in BODS?** _[Summarise these needs. Link to user stories, reports, blogs and other evidence where possible. Or add user stories here directly.]_

The [Open Ownership Principles](https://www.openownership.org/principles/comprehensive-coverage/) suggest that details of PLC intermediaries should be disclosed in beneficial ownership declarations; also, that where PLCs are exempt from disclosing their beneficial owners, they should declare the company's identity and the basis of the exemption. 

Publicly listed companies (PLCs) are often exempt from full beneficial ownership disclosure (on the basis that information about their ownership and control should already be public). Even if this information can be discovered from public filings and documents, there is a need to unambiguously identify PLCs in BODS data and facilitate the discovery and linking of related datasets. 

Open Ownership released a technical note, *[Beneficial Ownership and Listed Companies](https://www.openownership.org/uploads/technical-note-beneficial-ownership-and-listed-companies.pdf)* with advice on when and how to require disclosure of details of a publicly listed company, for the sake of beneficial ownership transparency. The Appendix of the document sets out what information should be published about PLCS.

**What impact would not meeting these needs have?**
If BODS could not hold identifying and useful information about PLCs:

i) The information would be collected and published in non-standard ways
ii) Insufficient information might be collected for the purposes of identifying PLCs
iii) Complete pictures of beneficial ownership would be more difficult to uncover in particular cases.

**How important is it to meet the above needs?**

Very. Publishing details of public companies' listings is an integral part of transparency of beneficial ownership, as reflected by the Open Ownership guidance linked to above. 

**How urgent is it to meet the above needs?**

[...]

PR #361, opened 2021-09-23
Add Feature template to issue templates

kd-ods:
This is a new issue template based on that used for these feature tickets:

SOEs: https://github.com/openownership/data-standard/issues/360
PLCs: https://github.com/openownership/data-standard/issues/359


StephenAbbott (approved):
Nice. Thanks @kd-ods. Approving with a couple of tweaks:

1. Instead of talking about a deficit in the standard (which felt too negative 😄), I've changed the intro language to say "where changes or revisions to the standard may be required".
2. Feature name is now Heading 2

Issue #264, opened 2019-09-19
Remove Markup from schema descriptions


rhiaro:
Inline markup like hyperlinks is currently included in some `description` fields in the schema. So far, markdown has been used. There's no guarantee or expectation that the schema JSON will be rendered by something that can parse markdown, and indeed on our own docs markdown is only (at the time of writing) displayed correctly in the tables that are rendered by the `jsonschema` widget, and not by Sphinx directly, resulting in bugs like:

![Screenshot at 2019-09-19 16-15-37](https://user-images.githubusercontent.com/466229/65252271-f47ba800-daf8-11e9-8d8d-6a23a426ba87.png)

The JSON descriptions should really be plaintext strings, as they don't have a media-type indicator of any kind to say how they should be treated otherwise.

I propose we remove all kinds of inline markup from the schema descriptions. If hyperlinks are needed, they can simply be written out, eg. "..based on the EC ISA Core Person Vocabulary (https://joinup.ec.europa.eu/solution/e-government-core-vocabularies) definitions."

If in agreement, the [style guide](https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md) should also be updated to note this.

Related: https://github.com/openownership/data-standard/issues/200

timgdavies:
I lean towards allowing markdown links in plaintext where there is a reasonable chance the text will be rendered by a markdown parser (and would instead suggest we look at whether we can get docson support for markdown working here). 

My reasoning is that: 

* Markdown links are specifically designed to be human readable. The only different from how text would otherwise be written is the square brackets;
* If makes for more user friendly documentation if prose is not interrupted by long URLs;

However, happy for discussion on this.

odscjames:
There is other issues on this and a different approach was taken before. I'll try to find links. Basically we had a problem where the schema markdown was shown in 2 widgets. One could handle the MarkDown and one could not. We decided the correct path here is to not have MarkDown in the schema and instead where we need to do nice display things, we do special things in the widget.

> where there is a reasonable chance the text will be rendered by a markdown parser 

Our feeling at the time was that we can't say there is a "reasonable chance the text will be rendered by a markdown parser ". A schema may be taken out of the standard and re-used in any number of tools, and the standard designers could easily have no knowledge of that. Therefore the schema files have to be a reusable as possible.


(Also I don't think this is related to #226, these are using Markdown in 2 different things and having recommonmark installed or not does not effect these display widgets.)

timgdavies:
[...]

PR #370, opened 2021-10-20
Remove markdown from schema descriptions

rhiaro:
Removes markdown from schema descriptions per #264 and replaces missing backticks and quotes inline with style guide.

### DIFF
--- a/schema/components.json
+++ b/schema/components.json
@@ -100,7 +100,7 @@
         },
         "scheme": {
           "title": "Scheme",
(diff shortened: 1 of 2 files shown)


kd-ods (approved):
Above and beyond, Amy. You've actually improved the punctuation along the way :-)

(And I'm assuming from there only being 2 files with changes that there was no markdown to strip from the person or ownership-or-control schemas.)

Issue #318, opened 2020-06-04
Move to https://pypi.org/project/sphinxcontrib-opendataservices-jsonschema/


odscjames:
For jsonschema widget, we have been using a branch of a forked copy. We want to clean that up into using a properly published and maintained library. This will ensure we get up to date code in the future and other new features. 

But some work needs to be done on that lib first - see https://github.com/OpenDataServices/sphinxcontrib-opendataservices-jsonschema/issues/24

Issue #226, opened 2019-07-10
Remove recommonmark dependency from docs


rhiaro:
Since all of the documentation pages are now RST we no longer need a markdown parser in sphinx.

TODO: Triple check for any stray markdown, eg. there are still some markdown links in schema descriptions at the time of writing. Not sure if these are rendered to HTML by Sphinx or docson/jsonschema though.

timgdavies:
Related to #264

odscjames:
I don't think this is related to #264 - see https://github.com/openownership/data-standard/issues/264#issuecomment-617784175

odscjames:
We can remove this from our explicit requirements but to fully remove this we need to also upgrade:
* sphinxcontrib-jsonschema / sphinxcontrib-opendataservices-jsonschema to 0.2 or above and this is not simple, it depends on #318
* sphinxcontrib-opendataservices to 0.2 - this actually depends on sphinxcontrib-opendataservices-jsonschema so again #318 is a dependency

PR #378, opened 2021-11-04
requirements: Python 3.8, upgrade, remove recommonmark, pypi releases

odscjames:
https://github.com/openownership/data-standard/issues/318
https://github.com/openownership/data-standard/issues/226

Start using pypi releases of our libraries.

Have to upgrade Sphinx as myst-parser needs a newer version

Python upgraded needed.
Python 3.8 is the latest version RTD seems to support at this time

theme_overrides.css:

Add CSS classes to get back old behavoir changed by upgrade.

Remove wy-table-responsive css classes.
I couldn't find any place they were being used.
Also the CSS file wasn't loaded before so they were not making a
difference.

Should be no changes to content, so no changelogs, translations, etc.

### DIFF
--- a/docs/_static/theme_overrides.css
+++ b/docs/_static/theme_overrides.css
@@ -1,31 +1,10 @@
-/* override table width restrictions */
-.wy-table-responsive table td, .wy-table-responsive table th {
-    white-space: normal !important;
+
+/** A newer version of docutils aligns center by default. We want the old behavior of left align. **/
+body .align-default {
+    text-align: left !important;
 }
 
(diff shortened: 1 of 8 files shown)


odscjames:
https://standard.openownership.org/en/226-python-deps/

rhiaro:
I can't see any i18n problems as a result of upgrading sphinx. All looks fine so far.

odscjames:
Did CSS without changing the theme

rhiaro (approved):

PR #390, opened 2021-12-30
Remove 0.0.3 warning banner

rhiaro:
Removes manual old version warning banner from 0.0.3 version of docs (as one is automatically provided by readthedocs). For https://github.com/openownership/data-standard-sphinx-theme/issues/67

### DIFF
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -1,11 +1,6 @@
 Conformance and validation
 ===========================
 
-```eval_rst
-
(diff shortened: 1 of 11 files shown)


kd-ods (approved):
Looks good to me. (I can see on the live site that the RTD banner is working and that removing this custom text can be done safely. https://standard.openownership.org/en/0.0.3/provenance.html)

PR #391, opened 2021-12-30
Remove 0.0.2 warning banner

rhiaro:
Removes manual old version warning banner from 0.0.2 version of docs (as one is automatically provided by readthedocs). For openownership/data-standard-sphinx-theme#67

### DIFF
--- a/docs/credits.md
+++ b/docs/credits.md
@@ -1,11 +1,4 @@
 Credits (stub)
 =======
 
-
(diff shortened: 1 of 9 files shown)


kd-ods (approved):
Again, I can see this will sort out the double-message here: https://standard.openownership.org/en/0.0.2/examples.html)

PR #393, opened 2022-01-06
Add not latest version warning to 0.1.0

rhiaro:
Uses a version of the theme with the "not latest version" warning, for https://github.com/openownership/data-standard-sphinx-theme/issues/67. See also https://github.com/openownership/data-standard-sphinx-theme/pull/74

This references a different data-standard-sphinx-theme commit to that on the main branch as there have been updates to the theme since 0.1.0 was released, so I've inserted the necessary update on a new branch (https://github.com/openownership/data-standard-sphinx-theme/tree/0.1.0-theme) without those updates to avoid breaking old version docs with these or future additional theme changes. It does mean we need to make sure that branch is not deleted in future (we should probably consider versioning the theme in the medium term though). 

I did also have to update another dependency as the older version wouldn't build on readthedocs. This doesn't seem to have affected the 0.1.0 build of the docs, but a review for this would be useful to double check everything works as expected. Preview here: https://standard.openownership.org/en/0.1.1/


kd-ods (approved):
I've had a browse and it looks like everything is working as expected.

Issue #363, opened 2021-09-30
Implementation proposal: Represent State-Owned Enterprises (SOEs) no.1


kd-ods:
_[This ticket proposes a particular approach to implementing a feature in BODS. In most cases, a proposal should be made only after the problem it tries to solve is described in a 'feature' ticket._

_See [Feature development in BODS](https://openownership.github.io/bods-dev-handbook/feature_development.html) in the Handbook._

_Comments on this ticket can be used to question, refine and develop this implementation proposal. Interactions and work on this ticket represent a collaborative process. Proposals may be paused, withdrawn, or developed into draft implementation plans. Update the 'Proposal status' as thinking progresses on this thread. Highlight changes or updates to the proposal within thread comments, with a clear 'Updated proposal' heading.]_

## Implementation proposal for: Represent State-Owned Enterprises (SOEs)

**Feature ticket link** #360

Implementation proposal status: **ACTIVE** (Not: paused | withdrawn | being implemented)

### Initial proposal

_[Broadly, what changes do you propose are made to the BODS schema or documentation in order to develop the feature in question.]_

**Overview**

Given the needs and constraints set out in the feature ticket, #360, this proposal is that: states and state bodies are represented by types of Entity Statement. They can then be connected in the standard way via Ownership-or-control Statements with other entities. Companies and other entities will not be flagged (within their statements) as being ‘state-owned’. Rather, their connections within the data to states or state bodies will be evident.

This allows BODS to (1) remain un-tied to any particular definition of ‘state-owned enterprise’, and (2) represent the gamut of ways in which states, directly or indirectly, own and control enterprises.

**Proposed changes to the schema**

1) To the [entityType codelist](http://standard.openownership.org/en/0.2.0/schema/reference.html#entitytype) we’d add ‘state’ and ‘stateBody’ types.

2) In [Entity Statement](http://standard.openownership.org/en/0.2.0/schema/reference.html#entitystatement), add a new `entitySubtype` property which takes an object with `general` (enum) and `local` (string) properties. The value of `general` would be taken from a stateBodyType codelist:

- Government department
- State Agency
- Other

The value of `local` would provide the locally-used descriptor for the state body type.

3) In Entity Statement, change 'Incorporated In Jurisdiction' property to just 'Jurisdiction'. (And change the description to 'The jurisdiction in which this entity was registered (for legal and registered entities, and arrangements). Or the state's jurisdiction (for states and state bodies).')

[...]

Issue #162, opened 2019-04-30
Reduce the complexity of approximate date fields by dropping some of ISO8601


stevenday:
Date fields that allow approximate values have a hugely complex regex listed in the schema docs for their validation. In writing BODS import code, I had to break this apart piece by piece to understand what kind of data I might be getting and I was a bit surprised by all the options it allowed.

In particular, I wonder whether we could drop the following:
- Times and timezones (all of the fields I could see are for *dates*, where a time wouldn't really add much useful data)
- Week numbers
- Ordinals (days of the year)
- Optional separators

This would leave it just allowing YYYY, YYYY-MM or YYYY-MM-DD (I think, I'm still not 100% sure if just a year is ok by the current regex). 

My rationale for this is that people are most likely going to be formatting their date/time data into ISO8601 anyway (from a local format, timestamp, etc), so I don't think the extra flexibility of these options is really helping them. On the contrary, by making the standard simpler and more explicit, I think it will help both publishers and consumers.

ScatteredInk:
Agree with this - current regex is unhelpful and this would be a relatively easy fix.

timgdavies:
See #12 for some background on dates. 

Research question for @stevenday and @kd-ods: Are we always seeing complete dates in incoming data? Or is handling incomplete / fuzzy dates showing up from experience as an important thing to consider?

stevenday:
> Research question for @stevenday and @kd-ods: Are we always seeing complete dates in incoming data?

To the best of my knowledge all of the register's sources provide complete dates, so I'm minded to say it's not a very high priority at the moment.

siwhitehouse:
We have a practical instance of the impact of including time in some fields. When we put example data through the Data Review Tool, it is appending 'T00:00:00+00:00' to the date in the following fields (at least):

- birthDate
- foundingDate
- startDate

As we are about to share this document with the overseas territories should we pick this off as a piece of work to do asap? If so, am I correct in thinking that we need to amend the regex? What else, if anything, needs to be done?

Bjwebb:
> When we put example data through the Data Review Tool, it is appending 'T00:00:00+00:00' to the date in the following fields (at least):

I think this is only happening for spreadsheets. If so, its not quite being appended, but is an artifact of how we read the underlying date time information (different to how it is displayed) in flatten-tool. We can change what flatten-tool does, but it may have implications for exactly what change you want to make to the schema. Happy to have a brief chat about this.

kd-ods:
[...]

PR #403, opened 2022-02-16
Representation of SOEs in BODS data

kd-ods:
# Overview

Adds ability to represent SOEs in BODS data. See changelog for full details.

The main changes are in entityStatement.json and in the codelists. I've added 3 new test files too. 

I had to update lots of test and example data because of renaming incorporatedInJurisdiction to jurisdiction.

The other thing to be aware of is that I've gone for the easy option in terms of modelling the relationship between entityType and entitySubtype.generalCategory. There's no json schema validation involved, and I've built in the entityType code into the entitySubtypeCategory codes. It's repetitive but should help usability. There may be better options.

Jack, I've asked for your review since you've been involved with the SOE modelling from day 1. David - I thought your eyes on it could be useful, if you need a break from other bits at some point.

(Closes/Relates to) issue: #363 

## Translations

- n/a -  I've notified the translation coordinator of any new strings that will need

## Documentation & Release

- [x ] I've thought about how and when this needs to be released. See:
- [x ] I've updated the changelog, if necessary.
- n/a I've updated [reference.rst](https://standard.openownership.org/en/latest/schema/reference.html) to reflect any changes to schema structure or naming.
- **I'll do this as a later PR**...I've added or edited any other related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md


schema/entity-statement.json
```diff
@@ -166,6 +168,53 @@
...
+        "name": {
+          "type": "string",
+          "title": "Statute name",
+          "description": "The name of the law. "
+        },
+        "date": {
+          "type": "string",
+          "title": "Date",
+          "description": "The date on which the law was passed. The date SHOULD be in the form YYYY-MM-DD. When only the year or year and month is known, these can be given as YYYY or YYYY-MM.",
+          "pattern": "^([\\+-]?\\d{4}(?!\\d{2}\b))((-?)((0[1-9]|1[0-2])(\\3([12]\\d|0[1-9]|3[01]))?|W([0-4]\\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\\d|[12]\\d{2}|3([0-5]\\d|6[1-6])))([T\\s]((([01]\\d|2[0-3])((:?)[0-5]\\d)?|24\\:?00)([\\.,]\\d+(?!:))?)?(\\17[0-5]\\d([\\.,]\\d+)?)?([zZ]|([\\+-])([01]\\d|2[0-3]):?([0-5]\\d)?)?)?)?$"
```
ScatteredInk, line 214: This validates "2011-12-03T10:15:30Z" etc. The SHOULD is a bit ambiguous for me here given that only three possibilities are given and the validation is more permissive than the text and the pattern is unreadable for anyone not familiar with what the regex does :)

ScatteredInk (changes requested):
@kd-ods This looks great - thanks for your work on this. I think it would be good to clarify the codelist description of legal entity to avoid ambiguity. I'm ambiguous about the date pattern: I don't think it's ideal that it's so obscure but I think it really just adds weight to resolving #162

schema/codelists/entityType.csv
```diff
@@ -1,6 +1,8 @@
 code,title,description,technical note
 registeredEntity,Registered Entity,"Any legal entity created through an act of official registration. In most cases, registered entities will have an official issued identifier.",
-legalEntity,Legal entity,"A body with distinct legal personality, such as a government department or international institution, but which is not otherwise uniquely identified in some official register.",
```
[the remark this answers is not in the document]
kd-ods: @ScatteredInk - I removed, 'such as a government department' from the legalEntity definition. But maybe the addition of the bolded words below would add more distinction:

Legal entity,"A body with distinct legal personality, such as an international institution **or statutory corporation**, but which is not otherwise uniquely identified in some official register."

State body,A **core** administrative or legislative unit within a state's apparatus.,

I think that helps a bit? (There will always be edge cases that sit on the boundary between definitions. But we should aim to make that a narrow boundary!)

FWIW I'd say the Oil & Pipelines Agency is a legal entity.

What do you think?

schema/entity-statement.json
kd-ods: I agree - it's not ideal at all. This non-optimal pattern replicates what is currently in the schema for similar dates.  My preference is to do a thorough job via #162, though, in a future release.

ScatteredInk: Agree - that sounds like the best approach.

schema/codelists/entityType.csv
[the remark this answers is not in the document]
ScatteredInk: Yes, that sounds good to me.

ScatteredInk (dismissed):

ScatteredInk (approved):

Issue #299, opened 2020-03-11
Translation framework: duplication of translation of codelists text


kd-ods:
Reviewing a [PR on the Spanish translation](https://github.com/openownership/data-standard/pull/295) it became clear that there's an inefficiency in the way the strings are pulled out and represented as resources in Transifex (pot files?). Codelists are being translated twice. 

The schema-reference.po contains the all the codelist strings, as well as the codelist strings being represented in the codelists.po file.

The translation pipeline needs to be rationalised.

jpmckinney:
The `csv-table` directive will yield the strings in the CSV for translation.

To get around this, OCDS uses at times the `csv-table-no-translate` directive from sphinxcontrib-opendataservices or the `codelisttable` directive from https://github.com/open-contracting/sphinxcontrib-opencontracting (the latter renders Markdown in codelist descriptions).

For those directives to work, OCDS needs to translate the codelist files that are referenced from its `codelists.md` page. It does that here: https://github.com/open-contracting/standard/blob/1.1-dev/docs/conf.py#L138

PR #409, opened 2022-03-02
docs: don't translate embedded codelists

rhiaro:
# Overview

Excludes codelist tables in the Schema Reference docs from being exported for translation, as they are redundant.

Run `tx pull -a` (to get latest from transifex), then `cd docs` and `make gettext` to extract strings from the documentation. Find the local file `_build/gettextschema/reference.pot` and verify that no strings directly from the codelists are in there (ie. no strings from `../../_build_schema/codelists/*`).

What I haven't done is pushed this up to transifex, since it's currently hooked up to v0.2 and I think that's frozen at the moment, so the redundancy remains on transifex and in the translated `.po` files for the time being.

Anyone familiar with the translations workflow.

(Closes/Relates to) issue: #299

## Translations

- [n/a] I've notified the translation coordinator of any new strings that will need

## Documentation & Release

- [n/a] I've thought about how and when this needs to be released. See:
- [n/a] I've updated the changelog, if necessary.
- [n/a] I've updated [reference.rst](https://standard.openownership.org/en/latest/schema/reference.html) to reflect any changes to schema structure or naming.
- [n/a] I've added or edited any other related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -19,7 +19,6 @@ source_file = docs/_build/gettext/svg.pot
 source_lang = en
 type = PO
 
-
 [bods-v02.index]
 file_filter = docs/locale/<lang>/LC_MESSAGES/index.po
 source_file = docs/_build/gettext/index.pot
--- a/docs/schema/reference.rst
+++ b/docs/schema/reference.rst
@@ -301,7 +301,7 @@ Codelists
 AddressType
 +++++++++++
 
-.. csv-table::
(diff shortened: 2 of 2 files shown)


siwhitehouse (approved):
Thanks @rhiaro. I followed your instructions and it all works as described.

Issue #401, opened 2022-02-16
CSV tests apply to built schema only - as expected?


kd-ods:
I made a change to the entityType enum in entityStatement.json, adding a couple of new codes. I also added these in the related codelist. 

When I ran pytest, I got a fail: the new enums in the entityStatement.json weren't being 'picked up'. Confusing. Looking at the error output:

> data-standard/tests/test_json.py::test_codelist_enums[/home/kadie/code/data-standard-sphinx-theme/data-standard/docs/_build_schema/entity-statement.json-entity-statement.json-data0]
>   Codelist mismatch:
>   
>                       entityType.csv: 
>   
>                       ['registeredEntity', 'legalEntity', 'arrangement', 'anonymousEntity', 'unknownEntity', 'state', 'stateBody']
>   
>                       entity-statement.json enum:
>   
>                       ['registeredEntity', 'legalEntity', 'arrangement', 'anonymousEntity', 'unknownEntity']

So the tests were looking in _build_schema for the entity statement (but not the codelist csv, apparently). Sure enough, when I rebuilt the docs, the codelists passed.

So - is this expected behaviour? Am I wrongly using pytest? 

Assigning to @rhiaro and @odscjames who might know.

rhiaro:
You're not doing anything wrong. It looks for all JSON files in the directory tree to compare against, but if it's not helpful to have it also check the built schema I can exclude that.

kd-ods:
> ...but if it's not helpful to have it also check the built schema I can exclude that

Yes - I think we should exclude that since, effectively, it's just testing whether the schema (or docs) have been built.

rhiaro:
This was fixed in https://github.com/openownership/data-standard/commit/0498c23fbba3c73bd535c8347e266fdfc11cbe82

PR #410, opened 2022-03-03
tests: only test source schema (ie. not locally built schema), for #401

rhiaro:
# Overview

Makes the schema/codelists tests more precise by only looking at changes to schema in the source schema directory, rather than scanning the whole data-standard directory for schema files. Previously, if someone had a built version of the schema in their local environment, it would pick these up and run the tests against them as well, which could cause failures if they hadn't been rebuilt since making changes to the sources (see #401).

Make a change to a codelist and corresponding schema enums and run the tests (`pytest tests/`).

Anyone familiar with schema edits and running the tests.

(Closes/Relates to) issue: #401

## Translations

- [n/a] I've notified the translation coordinator of any new strings that will need

## Documentation & Release

- [n/a] I've thought about how and when this needs to be released. See:
- [n/a] I've updated the changelog, if necessary.
- [n/a] I've updated [reference.rst](https://standard.openownership.org/en/latest/schema/reference.html) to reflect any changes to schema structure or naming.
- [n/a] I've added or edited any other related documentation. See: https://github.com/openownership/bods-dev-handbook/blob/master/style_guide.md

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -33,7 +33,7 @@ def test_invalid_json():
     )
 
 
-schemas = [(path, name, data) for path, name, _, data in walk_json_data() if is_json_schema(data) and not path.endswith('tests/schema/meta-schema.json')]
+schemas = [(path, name, data) for path, name, _, data in walk_json_data(top=absolute_path_to_source_schema_dir) if is_json_schema(data) and not path.endswith('tests/schema/meta-schema.json')]
 with open(os.path.join(this_dir, 'schema', 'meta-schema.json')) as fp:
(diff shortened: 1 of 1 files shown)


kd-ods (approved):
Yes - this works. I tested it by:

1) Running the tests on the clean branch: all passed

2) Adding a codelist item to a csv, but not to the related enum.

3) Ran the tests again: fail (as expected)

4) Built the docs

5) Added the codelist item to the enum to align everything.

6) Ran the tests again: pass. (Previously, this was unexpectedly failing because of stage (4))

Issue #222, opened 2019-07-03
Internationalisation: some strings not translated/translatable in schema tables


kd-ods:
See: http://standard.openownership.org/ru/0.2-dev-ru-translation/schema/reference.html#publicationdetails

The values under the last three columns (Type, Format, Required) are not translated. So we have 'string', 'object', 'date', 'uri', 'Required' etc. Now this might be acceptable, since these are keywords in JSON schema.

However we also have 'See' (as in 'See Publisher') left untranslated. That should be translated.

odscjames:
I will want to have a quick chat with @rhiaro about the translation process when she has time ...

siwhitehouse:
There are some more places where we have non-translated strings, notably in some of the headings and navigation menu strings. I've completed the import of the sphinx-theme strings today (many thanks to @odscjames) and I'll detail them on Monday.

siwhitehouse:
See http://standard.openownership.org/ru/0.2-dev-ru-translation/index.html where the headings "For Newcomers" and "For Developers" have not been translated.

Checking in https://www.transifex.com/OpenDataServices/bods-v02/translate/#ru/about--index/174975439 shows that this did not get pushed up to Transifex (assuming that is the correct file).

Checking locally in the docs/locale/ru/LC_MESSAGES/index.po file I don't see either of these headings.

odscjames:
I separated the front page issue into https://github.com/openownership/data-standard/issues/223

rhiaro:
'string', 'object', 'date', 'uri', shouldn't be translated as these are JSON keywords.

In the context of the docs, 'Required` could be. I spent an hour hunting through all of our own templates, as well as the [jsonschema](https://github.com/lnoor/sphinx-jsonschema) and Sphinx source code and I can't find where this "Required" string, or "See" come from though. Most useful clue I found on my treasure hunt is https://github.com/lnoor/sphinx-jsonschema/blob/master/sphinx-jsonschema/wide_format.py#L73 possibly, for "See"? I think it then gets handed to Sphinx templating. To be continued.

rhiaro:
Update: @odscjames found the offending line, I was looking in completely the wrong place: https://github.com/OpenDataServices/sphinxcontrib-jsonschema/blob/openownership/sphinxcontrib/jsonschema.py#L142

siwhitehouse:
Am I correct in thinking that we have identified where the issue lies, but not necessarily how to fix it?

rhiaro:
[...]

PR #431, opened 2022-04-05
i18n: re-lable codelist links to avoid need to translate

rhiaro:
# Overview

Links to codelists from the schema reference were labelled "Codelists" in such a way that could not be extracted for translation. I have relabelled these to use the name of the codelist, which then doesn't need translation. 

Towards: https://github.com/openownership/data-standard/issues/222

[Preview on readthedocs](https://standard.openownership.org/en/rhiaro-222-translate/schema/reference.html)

## Translations

## Documentation & Release


siwhitehouse (approved):
I see all of the changes as expected

PR #432, opened 2022-04-06
[0.2.0] translations: sync with transifex

rhiaro:
Some wrong source files accidentally got pushed to the Transifex 0.2.0 project a little while ago, so this PR re-syncs things.

There remains one string that is not translated for v0.2, on the About > Privacy Notice page. I can't find an old translation for it anywhere in either language.

Also deleted the FR locale directory as this was never translated.

Preview [RU](https://standard.openownership.org/ru/0.2.0-transations-sync/about/index.html) and [ES](https://standard.openownership.org/es/0.2.0-transations-sync/about/index.html) to verify translations are still intact. In particular check About (index) and About > Credits.


docs/locale/ru/LC_MESSAGES/about/credits.po
```diff
@@ -62,7 +62,7 @@ msgstr ""
 "<http://www.opendataservices.coop>`__), при этом было получено содействие "
 "рабочей :doc:`группы по стандарту данных <governance>`."
 
-# 51628ca7072b4f529336319cecfae3dc
+# e739e64e46cb49629b1cb7e80cfbc2dc
```
siwhitehouse, line 65: l78 is presented as 

Версию 0.2 написали Джек Лорд, Саймон Уайтхаус и Кади Армстронг ([`Open Data Services Co-operative http://www.opendataservices.coop>`_](https://standard.openownership.org/ru/0.2.0-transations-sync/about/credits.html#id7)),

in my browser (Vivaldi on Ubuntu), ie it shows the url rather than a hyperlink.

siwhitehouse (commented):
There is just one nagging hyperlink that is broken.

That outstanding privacy-notice looks like it was updated in August 2020, which is after the 0.2.0 translation work, so it is understandable why that has no translation.

https://github.com/openownership/data-standard/blob/0.2.0/docs/about/privacy-notice.rst

docs/locale/ru/LC_MESSAGES/about/credits.po
rhiaro: I pushed a fix for this but it seems to have lost the RU translation altogether :roll_eyes: Investigating..

PR #434, opened 2022-04-06
Fix docs build warnings

rhiaro:
# Overview

Fixes some minor bugs in the documentation:

* Headers on About page used CSS instead of HTML for subheadings (not accessible, and inconsistent with rest of docs).
* Link to `support` from Schema > Concepts went to a page that [had been deleted](https://github.com/openownership/data-standard/commit/29279b163c08cbcaa78d65ae504796380cc852a5#diff-8054c3a0a30617900afb4c82a08bf99357c72c8aeaa89255d0bc5066792580f8) so changed this to link to Contact section on About page (which has the same text).
* Stray space in Use Cases markup.

## Translations

## Documentation & Release

### DIFF
--- a/docs/about/index.rst
+++ b/docs/about/index.rst
@@ -6,10 +6,8 @@ This website contains the specification of `version 0.3 of the Beneficial Owners
 
 BODS is an open standard providing guidance and a structured template for collecting, sharing and using high-quality, machine-readable data on beneficial ownership.
 
-.. cssclass:: h2
-
-        Background
-
+Background
+++++++++++
 
 The utility of beneficial ownership data is enhanced when the data is available in a `structured format <https://www.openownership.org/en/principles/structured-data/>`__. This allows the data to be easily analysed and linked with other datasets, enhancing the data’s ability to expose transnational networks of illicit financial flows and support effective and timely due diligence.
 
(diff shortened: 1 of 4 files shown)


kd-ods:
Let me review this one. (I seem to remember using the css class directive as a workaround. Sthg to do with navigation maybe, Will build and check.)

kd-ods (commented):
@rhiaro - I'm hesitating to approve the about/index.rst edits. What the page gains in accessibility, the navigation loses in usability. The index page headings appear as top level menu items, but the other pages in the about section appear as sub-menu items:

**With standard rst headings**
image

**With cssclass workaround**
image

Maybe there's a sphinx navigation directive or something that would help here. So that we can get rid of the cssclass workaround, but still sort out the menus (so that only pages, not headings, appear in the About section of the docs).

rhiaro:
Thanks @kd-ods, didn't realise that, I'll look into it. The main motivation for adding the headings was so I could link directly to the Contact section, which had been moved to a subsection from its own Support page.

rhiaro:
@kd-ods is it any better if the subheadings appear, but the subpages appear as well on the same level? 

Screenshot at 2022-04-07 14-04-51

rhiaro:
@kd-ods I've updated it so the ToC on the left is how it used to be on the about page (with just links to sub pages), and the subheadings on the index page are now HTML. Confirmed the strings can still be extracted for translation, and that the link from Schema > Concepts to Contact still works.

kd-ods (approved):
It's good to go, @rhiaro :-)

PR #448, opened 2022-05-30
Update description of legal entity to match Open Ownership glossary

StephenAbbott:
Glossary updated on 30th May 2022. Pull request submitted here to update this in future BODS release https://docs.google.com/spreadsheets/d/1vdf6Emicrm6DiinmqXT6unpuEx__zbzO-x4UBzHfACM/edit#gid=1399019670

# Overview

- What does this pull request do? Updates the text description for a 'legal entity' in the EntityType codelist

(Closes/Relates to) issue: #

## Translations

## Documentation & Release

### DIFF
--- a/schema/codelists/entityType.csv
+++ b/schema/codelists/entityType.csv
@@ -1,6 +1,6 @@
 code,title,description,technical note
 registeredEntity,Registered Entity,"Any legal entity created through an act of official registration. In most cases, registered entities will have an official issued identifier.",
-legalEntity,Legal entity,"A body with distinct legal personality, such as an international institution or statutory corporation, but which is not otherwise uniquely identified in some official register.",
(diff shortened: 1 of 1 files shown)


kd-ods:
@StephenAbbott - when you say: "update this in future BODS release", do you mean after the BODS 0.3 release?

(I think getting this into v0.3 is prob more trouble than it's worth:

- The text itself will need some editing. The current description makes the point that a legal entity that is registered should be categorised as a registered entity. We'd need to retain that for the moment.
- We'd need to re-extract the strings and update Transifex.

We are going to be making changes to the entityType codelist before v1.0 anyway, so we could align with the glossary at that point.)

StephenAbbott:
@kd-ods Yes. I didn't mean that v0.3 release. Have communicated internally that we can't change this in v0.3 at this late stage

kathryn-ods:
closing but noted this change in the doc for rewrite of fields for 0.4

Issue #469, opened 2023-01-09
Upgrade to JSON Schema Draft 2020-12


lgs85:
BODS uses JSON Schema [Draft 4](https://json-schema.org/specification-links.html#draft-4). We should consider updating to the latest schema [Draft 2020-12](https://json-schema.org/draft/2020-12/release-notes.html). This issue is intended to be the start of a discussion rather than a detailed implementation proposal at this stage.

# Benefits

There are a number of reasons to upgrade the schema version. The below are illustrative examples and by no means comprehensive.

## Clarity around keywords
- `$id` replacing `id`, reducing potential confusion with instance properties
- `$defs` replacing `definitions`, for similar reasons
- `exclusiveMinimum` changed to represent a number (and replace `minimum`) instead of a boolean, reducing redundancy

## New features and keywords
- `examples` can be embedded into the schema
- `const` can be used to specify constant values
- `if` `then` and `else` can be used for [conditional application of subschemas](https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else)

## New formats

Many new formats, including:
- `date`
- `time`
- `email`
- `uuid`
- `iri`

# Considerations

Updating the JSON Schema version will require a substantial amount of work in addition to the schema changes themselves, potentially including:
- Updating documentation reference pages
- Updating the BODS validation tool
    - CoVE will work with JSON Schema 2020-12, but changes to keywords may cause additional checks to fail
    - Some validation that are currently additional checks may be able to be embedded within the schema itself
- Updating example data (e.g. to deal with changes to `exclusiveMinimum` and `exclusiveMaximum`

Proposals for new BODS features including change over time, separating core data from metadata, and representing declarations, will all involve substantial structural changes to the Schema. Using draft 2020-12 may help to facilitate some of these changes, and this is something that will require further research. In the meantime, it may make sense to consider an upgrade of the current schema as-is to JSON 2020-12.

lgs85:
I tried plugging the [entity statement schema](https://raw.githubusercontent.com/openownership/data-standard/master/schema/entity-statement.json) into alterschema and [got something](https://gist.github.com/lgs85/eee2102b384385821b56e53d8e159a80) that looks very sensible on first glance. Obviously this won't do any new/subjective stuff but looks like a very useful starting point.

duncandewhurst:
[...]

PR #485, opened 2023-07-31
Update JSON schema to 2020-12

rhiaro:
# Overview

Updates the base metaschema to JSON Schema 2020-12 (from draft-04) and corresponding fixtures and docs. Only updates to replicate existing functionality with the new JSON Schema version - does not use new features or improvements.

Also updates some testing dependencies and updates the tests accordingly.

Changes:

* `id` to `$id`
* `definitions` to `$defs`
* semantics of `exclusiveMinimum` and `exclusiveMaximum`
* `enum`s with one value are now `const`
* updated `meta-schema.json`

(Closes/Relates to) issue: #469

### Tasks remaining

* [x] Update bodsVersion numbers in examples and tests data
* [x] Update min/maxs in examples data
* [x] Work out why there's a test failing on github - it passes for me locally.

## Translations

  - The descriptions for `Interest / share / {minimum,maximum,exclusiveMinimum,exclusiveMaximum}` have changed

## Documentation & Release

  - See also: [PR to BODS dev handbook](https://github.com/openownership/bods-dev-handbook/pull/52)


rhiaro:
@kd-ods Please could you check you agree with the changes in https://github.com/openownership/data-standard/pull/485/commits/dae15d35d35ece4fb29f46f0d0e8d3df14682b42? 

Note that I have not made the same changes to any instances of `maximum` where there was also an `exact` value. This is a different meaning from in 0.3 (where `maximum` meant `exclusiveMaximum` by default) but I suspect in the cases with an `exact` value that was actually incorrect, and it should have been an inclusive maximum. Can you confirm that's also correct?

kd-ods (commented):
Right - I've eyeballed it all and suggested a couple of edits. I'm not the best person to review the tests and requirements though. I haven't thoroughly reviewed the metaschema, either: if you think I ought to be able to review it (for the sake of improving my schema development skills!) then let's go through it together next week?

radix0000:
Have a couple of questions, but they aren't major issues and wouldn't block merging:

* data-standard/tests/test_json.py: lines 18-22 - might it be better to use pytest fixtures for schemas and metaschema rather than global variables (but seems like is a more general issue in code, so even if good idea may not be worth it)?
* data-standard/tests/test_json.py: line 23 - does this commented out code need to be here for some informational reason?
* examples/: replacement of maximum with exclusiveMaximum etc., get the impression that in previous versions where exclusiveMaximum was a boolean, it defaults to false (though couldn't find a reliable reference), so is this subtly changing the example and is that what is wanted?

rhiaro:
Thanks @radix0000 

>    data-standard/tests/test_json.py: lines 18-22 - might it be better to use pytest fixtures for schemas and metaschema rather than global variables (but seems like is a more general issue in code, so even if good idea may not be worth it)?

+1 I've opened https://github.com/openownership/data-standard/issues/493 to capture this so we can work out how much work it is (probably not much?) and queue it for future improvement

>    data-standard/tests/test_json.py: line 23 - does this commented out code need to be here for some informational reason?

Removed

>   examples/: replacement of maximum with exclusiveMaximum etc., get the impression that in previous versions where exclusiveMaximum was a boolean, it defaults to false (though couldn't find a reliable reference), so is this subtly changing the example and is that what is wanted?

That's correct, the examples were technically wrong before, see https://github.com/openownership/data-standard/pull/485#issuecomment-1659640722

kd-ods (approved):
@rhiaro - this looks good, including your responses to @radix0000 review. I'll approve it but feel free to hold out for a dev's eye view from @radix0000.

PR #486, opened 2023-08-09
Schema: rationalize dates, fixes #162

### LINKED ISSUE #162: Reduce the complexity of approximate date fields by dropping some of ISO8601

rhiaro:
# Overview

Simplifies dates as detailed in #162

- Interest `startDate` and `endDate`, PEP Status `startDate` and `endDate`, Entity `foundingDate` and `dissolutionDate`, and formedByStatute `date` are more strict - only date (YYYY-MM-DD) is valid; timestamps and partial dates are no longer valid.
- Person `birthDate` and `deathDate` are more strict - year (YYYY), year and month (YYYY-MM) and year, month and day (YYYY-MM-DD) are valid, but timestamps are no longer valid.
- Annotation `creationDate`, PublicationDetails `publicationDate` are less strict - valid as date or date-time (partial dates are still not valid).

Updates tests/date and adds a couple of extra tests.

Check the new schema descriptions and definitions against [the spreadsheet](https://docs.google.com/spreadsheets/d/1dQ1o0llbyxRSTCu0sX05AkGMB3RJEvzVmrJhswISv94/edit#gid=0)

@kd-ods 

(Closes/Relates to) issue: #162

## Translations

## Documentation & Release

### DIFF
--- a/schema/person-statement.json
+++ b/schema/person-statement.json
@@ -105,16 +105,30 @@
     },
     "birthDate": {
       "title": "Date of birth",
-      "description": "The date of birth for this individual. Please provide as precise a date as possible in ISO 8601 format. When only the year or year and month is known, these can be given as YYYY or YYYY-MM.",
+      "description": "The date of birth for this individual, in YYYY, YYYY-MM, or YYYY-MM-DD format.",
       "type": "string",
-      "pattern": "^([\\+-]?\\d{4}(?!\\d{2}\b))((-?)((0[1-9]|1[0-2])(\\3([12]\\d|0[1-9]|3[01]))?|W([0-4]\\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\\d|[12]\\d{2}|3([0-5]\\d|6[1-6])))([T\\s]((([01]\\d|2[0-3])((:?)[0-5]\\d)?|24\\:?00)([\\.,]\\d+(?!:))?)?(\\17[0-5]\\d([\\.,]\\d+)?)?([zZ]|([\\+-])([01]\\d|2[0-3]):?([0-5]\\d)?)?)?)?$",
+      "anyOf": [
+        {
+          "pattern": "^(\\d{4})(-(1[0-2]|0[1-9]))?$"
+        },
+        {
+          "format": "date"
+        }
+      ],
       "propertyOrder": 35
     },
     "deathDate": {
       "title": "Death date",
-      "description": "If this individual is no longer alive, provide their date of death. Please provide as precise a date as possible in ISO 8601 format. When only the year or year and month is known, these can be given as YYYY or YYYY-MM.",
+      "description": "The date of death for this individual, in YYYY, YYYY-MM, or YYYY-MM-DD format.",
       "type": "string",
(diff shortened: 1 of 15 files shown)


schema/person-statement.json
```diff
@@ -105,16 +105,30 @@
     },
     "birthDate": {
       "title": "Date of birth",
-      "description": "The date of birth for this individual. Please provide as precise a date as possible in ISO 8601 format. When only the year or year and month is known, these can be given as YYYY or YYYY-MM.",
+      "description": "The date of birth for this individual, in YYYY, YYYY-MM, or YYYY-MM-DD format.",
       "type": "string",
-      "pattern": "^([\\+-]?\\d{4}(?!\\d{2}\b))((-?)((0[1-9]|1[0-2])(\\3([12]\\d|0[1-9]|3[01]))?|W([0-4]\\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\\d|[12]\\d{2}|3([0-5]\\d|6[1-6])))([T\\s]((([01]\\d|2[0-3])((:?)[0-5]\\d)?|24\\:?00)([\\.,]\\d+(?!:))?)?(\\17[0-5]\\d([\\.,]\\d+)?)?([zZ]|([\\+-])([01]\\d|2[0-3]):?([0-5]\\d)?)?)?)?$",
+      "anyOf": [
+        {
+          "pattern": "^(\\d{4})(-(1[0-2]|0[1-9]))?$"
```
kd-ods, line 112: @rhiaro - is this definitely right? Testing it in a couple of online validators it won't match simple dates. Will "^([0-9]{4})(-(1[0-2]|0[1-9]))?$" do the job?

kd-ods (commented):
@rhiaro - Is \d definitely valid regex here? I can't get your pattern to match dates using https://regex101.com/ . Poss an issue with me or the validator!

Otherwise this PR looks ready to go to me.

schema/person-statement.json
rhiaro: `\d` is equivalent to `[0-9]`. The extra `\` in front is needed as an escape in the json schema pattern matching, but won't work if you paste it into online validators. This regex matches only yyyy and yyyy-mm, as the rest is covered by the `date` format in the `anyOf`. I've triple checked, this definitely works as expected.

kd-ods: Ah - it didn't occur to me that the escape `\` wasn't a regex escape but a json schema/python escape. I get it.

kd-ods (approved):

Issue #493, opened 2023-10-10
Test improvements


rhiaro:
From @radix0000 about improving the tests:

> use pytest fixtures for schemas and metaschema rather than global variables

rhiaro:
Complete in #510

Issue #487, opened 2023-08-09
BODS v0.4 schema redesign


kd-ods:
This ticket covers the work to implement the proposals on:

- #482
- #477
- #475

The work consists of:

- [x] Create new test and example data
- [x] Rework the json schema files. At a high level:
  - bods-package.json -> statement.json
  - entity-statement.json -> entity-record.json
  - person-statement.json -> person-record.json
  - ownership-or-control-statement.json -> relationship-record.json
- [x] Update existing test and example data
- [ ] Update docs and add 'To do' flags where pages will need ongoing work before the 0.4 release

jpmckinney:
And the links to the features for those proposals:

#463 
#465
#392

kd-ods:
@rhiaro - with the merging of https://github.com/openownership/data-standard/pull/510 , shall we call this ticket done?

(I'm satisfied that all the remaining work on related documentation, validation and fixing the docs build is captured in the remaining tickets on the [project board](https://github.com/orgs/openownership/projects/4).)

rhiaro:
@kd-ods I'm happy if you're happy. Still things to do / iterate on, but the fundamental changes have been made, and we're better off working to to tickets that represent specific smaller changes at this point.

kathryn-ods:
Noting that this feature is complete as of BODS 0.4 see - https://standard.openownership.org/en/0.4.0/standard/index.html

PR #510, opened 2023-12-12
0.4.0 schema redesign, JSON Schema version upgrade, and test refactor

rhiaro:
# Overview

  * Changes the schema to match the change over time, declarations, and metadata proposals (https://github.com/openownership/data-standard/issues/487)
  * Adds validation features now available in JSON Schema 2020-12 (#516)
  * Refactors the tests into schema structure tests (test_schema), schema validation tests (test_data) and example data tests (test_docs). Makes use of pytest fixtures for reusable components (#493).

  * Run the tests (`pytest tests`) - currently 28 tests in test_docs fail, because the old example data remains, which is all BODS version 0.3.0, but all other tests should pass.
  * Take data which is known to be valid 0.4.0 data, and covers the substantial semantic/structural changes that have been made to the schema, and put it in the tests/data/valid-statements directory, and run `tests/test_data.py` to confirm it validates as expected.

  * Kadie / Kathryn for the schema changes.
  * Ed for the tests refactor.

## TODOs

The following are not completed by this PR, but I don't think they need to block it at this stage.

* There are some todos remaining in the tests relating to bugs (all have open issues to track).
* The _invalid_ data tests are not complete - it's not equipped to check the correct error messages are being generated. 
* There is not enough test data to check all branches of the schema in data/valid-statements yet - we can keep adding this iteratively in smaller batches once this PR is merged.
* The old example data needs to be replaced - we can also do this in smaller batches.
* There are some further changes we can make for #516 which are documented in the spreadsheet, but which we can do as and when in smaller PRs.
* There's a test stubbed to check if the documentation builds - I'll add this at another point in future.
* The documentation fails to build - I'll fix this next, it might be a big job.

## Documentation & Release

- Discussed with Kadie about the changelog update.
- Documentation update is happening in other branches which will all eventually be merged into main with this.


tests/test_docs.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_examples_indent(examples_dir):
+    """
+    Check the data in the /examples directory are indented correctly.
+    """
+    errors = False
+    misindented = get_misindented_files(top=examples_dir)
+    for path in misindented:
+        warnings.warn(f'ERROR: Misintended files at {path}')
```
kathryn-ods, line 38: typo - misintended should be misindented

BibianaC: I have fixed the typo.
https://github.com/openownership/data-standard/pull/565

radix0000 (dismissed):
Have a few minor comments below but happy for PR to be merged, use best judgement on what might be useful to address or not:

General:

* There are lots of very useful doc-strings in code, but could be an improvement to add doc-strings with brief one line descriptions for all of the functions in tests, not just most of them, also there are some inconsistencies in use (e.g. comment at beginning of function where there are doc-strings in most other functions)
* There are some uses of multi-line strings at the global scope in modules (after first use as a module docstring), would these be better as comments in terms of readability/highlighting in editors etc.
* As noted elsewhere, there are a couple of typo/spelling issues e.g. "failling", "misintended" instead of "mis-indented" etc.

Formatting/Style:

* There is a commit focused on upgrading code to use f-strings rather than format, but there a still a few of uses of format in tests/test_schema.py:178,179 and docs/conf.py:273
* Inconsistent flake8 configuration: max-line-length is set to 119 in setup.cfg and the github workflow but there are number of lines longer than this: e.g. tests/test_schema.py:97, tests/conftest.py:237. This is due to flake8 being configured to ignore the E501 error in setup.cfg, which is rather inconsistent
* Some lines in tests/conftest.py could potentially be simplified/made more readable e.g. list comprehension that could just be a call to list(...) (131), unused loop variables that could be replaced with a _ (128, 247) etc.
* The default value of stacklevel (1) keyword is probably fine for warn calls in tests/test_docs.py, but it is best practice to explicitly specify it
* Obviously code could be run through black, increase consistency more e.g. import ordering etc.

kd-ods (approved):

PR #562, opened 2024-02-27
0.4.0 schema resdesign add test data

BibianaC:
Added valid test data. I haven't found anything else to add from the two examples.

Fixed Flake8 issues.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,7 @@ from jsonschema import FormatChecker
 from jsonschema.validators import Draft202012Validator
 from jscc.schema import is_json_schema, is_codelist, is_missing_property
 from jscc.exceptions import MetadataPresenceWarning
-from jscc.testing.checks import _false, _true, _traverse
+from jscc.testing.checks import _false, _traverse
 from jscc.testing.filesystem import walk_json_data, walk_csv_data
 from referencing import Registry, Resource
 from referencing.jsonschema import DRAFT202012
@@ -149,6 +149,7 @@ The fixtures in this file are automatically discovered by pytest without
 needing to import them in the test files.
 """
 
+
 @pytest.fixture
 def schema_dir():
     return get_schema_dir()
@@ -241,7 +242,7 @@ def codelist_json(request):
     `codelist_json` as a parameter with indirect=True, and the output passed to
     the test.
     """
-    codelist = request.param # (path, name, text, fieldnames, rows)
+    codelist = request.param  # (path, name, text, fieldnames, rows)
     return codelist[4]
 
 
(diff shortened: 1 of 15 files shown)


rhiaro (changes requested):
Thank you for fixing flake8!

The next step is to actually break this test file up even more - in the long run, I think it'll be easier to make sure every part of the schema is tested by having separate test data files for each field, eg. a file which tests `statementId`, a file which tests `birthDate`, a file which tests `names` only, and so on.. of course there will be a few more fields present in the file due to the required fields constraints, but the point is each file is the bare minimum to test _one_ particular field. That way it can be really obvious if something is missing and which files cover which fields (and would need to be changed if the constraints on a particular field was changed). Does that make sense?

BibianaC:
@rhiaro Thanks. Would this be what you mean by breaking the docs more?

**Example 1: Publication details**
```
[
        "publicationDetails": {
            "publicationDate": "2021-09-11T00:00:00+00:00",
            "bodsVersion": "0.4",
            "license": "http://opendefinition.org/licenses/cc-zero",
            "publisher": {
                "name": "Irish Business Authority"
            }
        }
]
```

**Example 2: Entity addresses**
```
[
        "recordDetails": {
            "addresses": [
                {
                    "type": "registered",
                    "address": "53 Killard, Kilteragh, Dooradoyle, Co. Limerick ",
                    "postCode": "V94 V3VT",
                    "country": "IE"
                }
            ]
        }
]

**Comment by rhiaro** [comment]
That's the right direction, but even more than that. So we can break down address in to files to test each individual address field, eg.

valid_person_address_type.json checks that the schema correctly identifies `registered` as a valid value of `type`.

```
[
        "recordDetails": {
            "addresses": [
                {
                    "type": "registered"
                }
            ]
        }
]
```

valid_person_address_address.json checks that the schema correctly validates the `address` field as a string.

```
[
        "recordDetails": {
            "addresses": [
                {

                    "address": "53 Killard, Kilteragh, Dooradoyle, Co. Limerick ",
                }
            ]
        }
]
```

valid_person_address_country.json checks that the schema correctly validates the `country` field as a string of length 2.

```
[
        "recordDetails": {
            "addresses": [
                {
                    "country": "IE"
                }
            ]
        }
]
```

In practice, there will need to be some other fields in the file for the whole thing to validate, because of the required fields on Statement (`statementId`, `declarationSubject`, `recordId`, `recordType`, `recordDetails`) and on Person (`personType`, `isComponent`) so the first example would look more like:

```
[{
        "statementId": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "declarationSubject": "xxx",
        "recordId": "xxx",
        "recordType": "person",
        "recordDetails": {
            "personType": "knownPerson",
            "isComponent": False,
            "addresses": [{
              "type": "registered"
            }]
        }
    }]
```

rhiaro:
You can see in conftest.py > `get_valid_data` function a bare minimum valid statement, and use that as a basis for adding in the fields to test specifically

rhiaro (approved):

Issue #558, opened 2024-02-20
Conditional validation of address types


kathryn-ods:
Valid address types are dependent on the type of statement. Can we validate this in the schema? 

**Person only**
placeOfBirth
residence
service

**Entity only**
registered
business

**Both**
alternative

kathryn-ods:
I'm going to experiment with this and see if I can get the validation working

kathryn-ods:
@kd-ods do you think we should restrict placeOfBirth to only being in the dedicated field? I think it's better to only have one place it can go.

kd-ods:
Yes, I agree - the 'placeOfBirth' code should only be used for that field.

PR #584, opened 2024-03-13
adding limitations on address types for different statement types

kathryn-ods:
# Overview

- What does this pull request do? limits the address types for different statement types. 
- How can a reviewer test or examine your changes? check that the tests run and edit some test data to have an invalid address type and check it flags an error (e.g. person can't have a business address) 

(Closes/Relates to) issue: #558

## Translations

## Documentation & Release


kathryn-ods:
@kd-ods @rhiaro this seems to work but we need to weigh up whether it's ok to have these additional restrictions defined within the person/entity schemas. These are reference out to an address object in components so for future changes we would potentially need to update the address object in components and also the address fields in entity/person which is a but clunky.

kd-ods (changes requested):
I've added some test data which passes and shows that the schema changes work. 

In person statements the Address object is also used for the placeOfBirth field and placeOf Residence field. So really we should check that the matching address type is used in those fields.

(Which then led me to think: why don't we just get rid of those two fields and people can publish the same information in the addresses array. But maybe that should be a consideration for a future slimming down of the standard.)

kathryn-ods:
Current preference is to remove placeOfResidence and placeOfBirth and have these within the addresses array. Going to sleep on this.

kathryn-ods:
Agreed with @kd-ods to remove placeOfResidence but leave in placeOfBirth. Separate ticket #604 for removing field. I'll look at adding type validation to placeOfBirth

kathryn-ods:
@kd-ods something odd is happening here - the tests are running successfully on my machine but failing here. I think it might be something to do with changes to the country code. Will investigate further.

rhiaro (dismissed):
I'm not completely convinced overriding the Address type where it's included as a $ref works like this - it might supplement, rather than override - but I haven't managed to find clear documentation of the behaviour. (If so, the solution might be to strip the type enum out of components.json altogether.) I'd prefer to test this a bit more, but won't have time in the immediate future, so happy for you to merge if you're getting the behaviour you expect, and we can revisit when we have more complete tests if they throw anything up later.

rhiaro (approved):

kd-ods (dismissed):
I can see there haven't been any more commits since @rhiaro reviewed, so approving now.

kathryn-ods:
@rhiaro something went a bit haywire here and @kd-ods and I couldn't figure out why. There's a bunch of changes from other branches showing up in the commit history, possibly due to a rebase. It won't allow a merge due to 'The merge-base changed after approval.' for now I am going to cherrypick into a new branch but maybe you can explain why this one went wrong

Issue #490, opened 2023-09-18
Guidance on statement ordering for BODS v0.4


kd-ods:
At the moment in BODS we state that:

> Each of the [entityStatements](https://standard.openownership.org/en/0.2.0/schema/reference.html#schema-entity-statement) or [personStatements](https://standard.openownership.org/en/0.2.0/schema/reference.html#schema-person-statement) referenced by a particular [ownershipOrControlStatement](https://standard.openownership.org/en/0.2.0/schema/reference.html#schema-ownership-or-control-statement) MUST appear before that particular ownership-or-control statement in the ordered array.

(From: https://standard.openownership.org/en/0.2.0/schema/guidance/serialization.html)

In BODS 0.4, relationship statements will connect to entity and person information via recordId, not via statementId. What are the implications of this for guidance on statement order in a dataset?

I think we still need at least one statement for a related record to occur before the recordId is used in a relationship record to support efficient processing. @radix0000 @rhiaro - does that seem right?

kathryn-ods:
@kd-ods noticed this is listed in in progress but doesn't have anyone assigned currently. Did you get an answer to this question elsewhere?

kd-ods:
Thanks, @kathryn-ods for the nudge.

No - this is still an open question. I've put it back into the 'to do' column.

@radix0000 @rhiaro - your opinions on this issue would be valuable.

kd-ods:
There also needs to be a similar requirement (to be added to the Serialization page) that: within a BODS dataset, the declarationSubject value of each Statement must match the recordId value of at least one other Statement.

This might have implications for statement ordering too: i.e. for more efficient processing Entity (or Person) Statements of declaration subjects must be the first Statements in a dataset.

But such stringent requirements on statement ordering may be extremely difficult for publishers to comply with. 

(There's also a requirement on statement ordering related to [use of componentStatementIDs](https://standard.openownership.org/en/latest/schema/guidance/repr-beneficial-ownership.html#intermediaries).)

kd-ods:
[...]

Issue #365, opened 2021-10-05
Clarify requirements around bodsVersion


kd-ods:
At the moment, there is nothing in the documentation about whether all statements in a BODS release/array should have the same BODS version. There is nothing about it in:

- [the bodsVersion description](http://standard.openownership.org/en/0.2.0/schema/reference.html#publicationdetails)
- [the serialization info](http://standard.openownership.org/en/0.2.0/schema/guidance/serialization.html)
- [the conformance and validation statement](http://standard.openownership.org/en/0.2.0/schema/conformance.html)

The [data review tool](https://datareview.openownership.org/) has to take a line on this - the line is that all statements in the submitted data should conform to the same BODS version. [Recent work](https://github.com/openownership/cove-bods/pull/75#pullrequestreview-771283947) provides better handling of instances where this is not the case.

Is there any reason why - for the next BODS release - we shouldn't update the docs to specify that all statements in a BODS release/array must have the same BODS version?

(@StephenAbbott @kindly @ScatteredInk @siwhitehouse)

StephenAbbott:
Is there a future risk here that BODS publishers who are early adopters of BODS v0.2 would move to BODS v0.3 once that is released and then the BODS release that they publish could contain a mixture of newer statements with bodsVersion 0.3 and older statements with bodsVersion 0.2? Or would all those statements likely be updated to bodsVersion 0.3?

kd-ods:
I *think* that historically-made statements which were still true/active should be updated to 0.3. That is: our advice to publishers would be that a data snapshot of _current_ BO information SHOULD all use the same version of BODS.

This is where we need to pin down our use of semantic versioning, I think. After, v1.0, it should be the case that: data published in conformance with a particular version of the standard will ALSO conform with future minor updates to the standard. So data published to v1.2 would also conform with v1.4.1 of the standard. There is a nice, [work-in-progress description](https://github.com/open-contracting/standard/pull/674/files) of what this means wrt the open contracting data standard.

It makes sense therefore to specify that in a BODS data release (= an array of BODS statements) all statements MUST have the same major version number. This is so that BODS-handling platforms can handle all statements in a release as if they conformed to the most recent BODS version referenced from a statement in the release. It also allows publishing platforms to adopt newer minor versions of BODS and publish to them, without having to convert or repackage older data.

kd-ods:
This can be addressed by editing the bodsVersion field description to read:

[...]

PR #600, opened 2024-03-25
Docs - Serialization page edit, closes #396 #490 #365

### LINKED ISSUE #396: Docs: add some info on encoding of BODS json
I don't think that any detailed guidance is required for the moment. But it's probably worth referencing the JSON specifications: IETF's [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259) and [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf). In particular, we might want to draw people's attention to:

- Section 9 of the ECMA-404 spec which deals with strings, encoding and escaping;
- Using UTF-8 for maximal interoperability (section 8.1 of RFC 8259);
- The option to escape characters that might cause problems when ingesting data (for example: &, <).

Is there anything else that should be covered?

I think [this page](https://standard.openownership.org/en/0.2.0/schema/guidance/serialization.html) could be edited to include the above.

kd-ods:
# Overview

Edits the Serialization page of the docs to address 3 tickets and update the content generally.

Build the docs. Read the page. 

(Closes/Relates to) issue: #

## Translations

## Documentation & Release


kd-ods:
@rhiaro - I'd appreciate your review on this. Especially because I wonder if the statement order requirements are too stringent.

kd-ods:
@radix0000 - see my question to @rhiaro above and the original issue #490. Let me know what you think of the requirements we're setting via this PR.

kd-ods:
I've taken out the new ordering requirement I'd added. I will add an issue about removing the old ordering requirement too and we can pick it up after 0.4.

PR #609, opened 2024-03-27
Remove placeOfResidence from person record

### LINKED ISSUE #604: Remove 'placeOfResidence' field
placeOfResidence doesn't need to be a separate field - it can just be in the addresses array

odscrachel:
# Overview

Removes `placeOfResidence` from the person record and creates invalid json file with the field included. NOTE this is currently failing the tests. I discussed this with Kathryn and we think it's due to the schema allowing extensions/extra fields, if so, then technically the invalid file is at the moment valid. 

Check whether we need to include invalid file or whether we need to update test/schema.

@kd-ods 

Closes #604 

## Translations

## Documentation & Release


tests/data/invalid-statements/inv_person_placeOfResidence.json
kd-ods, line 1: @odscrachel - I shouldn't have suggested you add this test file: I forgot that it would be treated as valid because we do [allow](https://standard.openownership.org/en/main/schema/conformance.html?highlight=extension) publishers to use 'custom' fields.

So you can just go ahead and remove this. And then the PR is good to go.

kd-ods (dismissed):
You can remove the test file - it's technically valid, but in the invalid-statements dir and causing the tests to fail. See my comment.

kd-ods:
@odscrachel - please get Kathryn to re-review this on Fri after you've made your edit.... then you can get it merged before hols :-)

kathryn-ods:
@odscrachel the pull_request test is failing because we merged in some changes yesterday with new test files and they don't have a statement date yet.

Not sure whether the best thing is to bring those changes into this branch or quickly update those test files on another branch. I need to learn about how to properly bring in changes from main.

odscrachel:
@kathryn-ods I've updated the relevant files now to allow the tests to pass for this branch

kathryn-ods (approved):

PR #620, opened 2024-04-02
Update error message in test_schema.py to point to json validation failure

kd-ods:
# Overview

Edits the test_schemas_loaded() error message to suggest the other reason why this test might fail. (As I discovered when I left some quotation marks in a schema description unescaped and caused half of all the tests to fail!)

(We could have a dedicated test to check that each of the schema files is valid json, but that's probably overkill.)

Eyeball and run pytest.

@rhiaro 

Closes #

## Translations

n/a

## Documentation & Release

n/a

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -44,7 +44,8 @@ def test_schemas_loaded():
     schemas_from_file = get_schema_paths()
     assert len(schemas_from_file) == len(
         schemas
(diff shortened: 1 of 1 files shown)


rhiaro (approved):
We *do* have a test to check the schema files are valid (`test_schema_valid`) (but this update is helpful if invalid json also causes this test to fail)

PR #624, opened 2024-04-08
Update names (person-record) to make fullName required

### LINKED ISSUE #536: Name object: descriptions and normative content need a re-think
The Name object description is :

> An name by which this individual is known. Names should be provided in `fullName`, and may optionally be broken down in the `familyName`, `givenName` and `patronymicName` fields, based on the EC ISA Core Person Vocabulary (https://joinup.ec.europa.eu/solution/e-government-core-vocabularies) definitions.

It would be preferable to edit the `fullName` description to say sthg like: "... a value SHOULD be provided for this property."

odscrachel:
# Overview

Makes fullName required in schema within names.
Updates existing valid test files
Creates invalid test file

@kd-ods or @kathryn-ods 

Closes #536 

## Translations

## Documentation & Release

### DIFF
--- a/schema/person-record.json
+++ b/schema/person-record.json
@@ -58,9 +58,9 @@
       "title": "Place of birth",
       "propertyOrder": 40,
       "$ref": "urn:components#/$defs/Address",
-      "properties":{
-        "type":{
-          "enum":[
+      "properties": {
+        "type": {
+          "enum": [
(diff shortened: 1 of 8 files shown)


kathryn-ods (approved):
All looks good - I think this has pulled in some changes from the addressType validation so we'll see what happens when it merges

PR #636, opened 2024-04-12
adding normative statement about source information

### LINKED ISSUE #537: Move old 'Sources and annotations' page content to 'Schema reference' page

The [information here](https://standard.openownership.org/en/latest/schema/guidance/provenance.html#provenance) would sit better as part of the [Schema reference](https://standard.openownership.org/en/latest/schema/reference.html) page. (There may be normative requirements that would better sit in the schema too.)

kathryn-ods:
# Overview

- What does this pull request do? adds a normative requirement that statements should include source info 

Closes #537 

## Translations

## Documentation & Release


kathryn-ods:
@kd-ods imo this is the only thing in the sources and annotations page that isn't otherwise covered in the schema/docs. See comments in #537 for more detail.

schema/statement.json
```diff
@@ -111,7 +111,7 @@
           },
           "source": {
             "title": "Source",
-            "description": "The source of information in this statement.",
+            "description": "The source of information in this statement. Each statement SHOULD contain source information",
```
kd-ods, line 114:
```suggestion
            "description": "The source of information in this statement. Each statement SHOULD contain source information.",
```

kd-ods (changes requested):
If we're deleting the provenance page, I feel we need to do something with the verification-related content... and also the point about using annotations to record data processing/cleaning/enrichment.

On the first, maybe just an edit to the `source.type` field description, to:

> "The types of the source, using the sourceType codelist. Include 'verified' in the array if the information in the Statement has undergone a verification process."

On the second, I actually think this is a bigger issue. If annotations are to be used to note data cleaning etc, then those reasons should be in the motivations codelist. BUT let's not get bogged down in that for this release. We can spin up a new issue with the title "Consider adding 'cleaning' and 'enrichment' to the motivations codelist" with some background and rationale. Then come back to it in the future.

What do you think?

kathryn-ods:
@kd-ods have made those changes and made #638 for the annotations issue

kd-ods (approved):
Great. This is good to go.

Issue #336, opened 2021-02-09
Simplify the entity type codelist


ScatteredInk:
We currently have five types of entity: `registeredEntity`, `legalEntity`, `arrangement`, `anonymousEntity`, `unknownEntity`.

 `registeredEntity` and `legalEntity` are a source of confusion due to conflicts with their common usage in other related standards, for example most legal persons under FATF guidance would be referred to as registered entities in BODS.

I suggest removing the `registeredEntity` category and aligning "legal" and "arrangement" categories (broadly) with the [definitions](http://www.fatf-gafi.org/media/fatf/documents/recommendations/pdfs/FATF%20Recommendations%202012.pdf) used by FATF for legal persons (R24) and legal arrangements (R25).

`legalEntity` (or `legalPerson`) would then cover registered companies _and_ the cases currently covered by `legalEntity`. Any additional details about the entity (for example, the fact that it is a government department with legal personality) can be covered through the proposed entity tagging system (currently #186 but full issue to follow) or by the use of identifiers. One advantage of using `person` rather than `entity` here is that our statements would become more generic   

@siwhitehouse As currently stated `arrangement` would cover the legal arrangements as described by FATF Rec 25 (which are broadly speaking express trusts and trust-like arrangements) _and_ any legal arrangements that give rise to ownership and control interests (like a revenue-sharing agreement or nominee shareholder agreement). We may want a separate term for these to allow for an easier distinction.  

(Note that other entries on this list may change due to work @rhiaro is doing on structuring missing info.)

kd-ods:
I think this proposal will work.

At the outer edges of the 'legal person' FATF definition are legal persons which are not - in our terms - registered entities. I'm thinking of unregistered clubs, societies and associations in the UK context, for example. If we just have 'legal entity' and 'arrangement' types then we'll need to be clear in the documentation under which of the two categories such unregistered bodies sit.

On the question of terminology -  'legal person' or 'legal entity'. Ooof. People with English as a first language can find the concept of natural vs legal personhood a challenge, so I think we should avoid it if we possibly can. BUT I do see why we need to consider it. Difficult one.

ScatteredInk:
## Proposal

Having thought about this a bit more, I would propose the following types:

**Legal entities**

This would cover legal entities as defined by FATF:

[...]

PR #644, opened 2024-04-16
adding in entity subtype validation logic and test files

### LINKED ISSUE #519: Validation changes - entitySubtype and entityType
In BODS 0.3 entitySubtype.generalCategory is only used for entities with entityType stateBody. 

In BODS 0.4 we are introducing an entitySubtype.generalCategory for trusts (#518) which will be valid for entityTypes arrangement and legalEntity. 

Currently this test https://github.com/openownership/lib-cove-bods/blob/master/libcovebods/tasks/checks.py#L622 checks the validity of an entitySubtype.generalCategory based on the prefix matching the entityType (e.g. stateBody-stateAgency) 

As 'trust' will be valid for multiple entityTypes we should:
- remove the prefixes from the codes
- find a new way to validate entitySubtype.generalCategory codes have the correct entityType

(Open contracting seem to have a system for this https://github.com/open-contracting/standard/blob/1.2-dev/schema/codelists/documentType.csv)

kathryn-ods:
# Overview

Closes #519

## Translations

## Documentation & Release

- [x] I've checked that the documentation builds without failing. See: https://github.com/openownership/data-standard#building-the-documentation-locally


schema/entity-record.json
```diff
@@ -54,7 +54,90 @@
...
+                  "stateAgency",
+                  "other"
+                ]
+              }
+            }
+          }
+        },
+        {
+          "if": {
+            "not": {
```
kd-ods, line 120: I think I'd just simplify this to: if 'registeredEntity' then 'other'.

We know that the entity codelist is likely to be simplified to remove 'registeredEntity' anyway in future, see #336 .

schema/entity-record.json
```diff
@@ -54,7 +54,90 @@
...
         }
       },
-      "propertyOrder": 4
+      "propertyOrder": 4,
+      "allOf":[
+        {
+          "if": {
+            "properties": {
+              "type": {
+                "const": "arrangement"
```
kd-ods, line 63: I'm slightly nervous about defining 'type' as an enum earlier and then using 'const' in these schemas for the logic. It works for immediate purposes, but I can imagine that it might cause problems with some json schema tools. Amy would know better then I.

To be on the safe side we can simply use "enum": ["singleValue"].

kd-ods (commented):
To test that the schema logic worked I added a couple of tests locally that you don't have here:

- a valid test which has an entitytype.type value, but no entitytype.subtype value
- an invalid test which has an entitytype.subtype value, but no entitytype.type value

So it's worth adding them on this PR.

Also, could you edit all the test filenames to include the field name, i.e "_type_" (so e.g. "entity_type_state_stateAgency"). It makes it easier to spot at a glance all the test fields for a particular field?

schema/entity-record.json
kathryn-ods: I can switch this to a positive rather than a negative but will also need to include unknownEntity/anonymousEntity/state

schema/entity-record.json
kathryn-ods: done!

schema/entity-record.json
kathryn-ods: have done this

kd-ods: Ahhh - yes. OK

kathryn-ods:
> * a valid test which has an entitytype.type value, but no entitytype.subtype value

Haven't added this explicitly because the majority of the valid entity files are formatted this way (e.g. entity.json)  - does it need a dedicated file?

> * an invalid test which has an entitytype.subtype value, but no entitytype.type value

For 'entityType.type is required' we'll presumably have a file with the type missing. Is it also useful to have a file with no type and a subtype? I've added one in but the error message is just that entityType is required. 

> Also, could you edit all the test filenames to include the field name, i.e "_type_" (so e.g. "entity_type_state_stateAgency"). It makes it easier to spot at a glance all the test fields for a particular field?

Done

kathryn-ods:
Agreed with Kadie we didn't need specific tests for a statement without a subtype, a statement with a subtype and no type. Added in test files for missing type and missing entityType as these are required fields.

kd-ods (approved):
Great :-)

tests/data/invalid-statements/expected_errors.csv
```diff
@@ -108,3 +108,16 @@ publication_details_publisher_no_publication_date.json,required,$[0].publication
 publication_details_publisher_no_bods_version.json,required,$[0].publicationDetails,bodsVersion
 publication_details_publisher_no_publisher.json,required,$[0].publicationDetails,publisher
 statement_declaration_not_string.json,type,$[0].declaration,declaration
+entity_trust_registeredentity.json,const,$[0].recordDetails.entityType.subtype,subtype
```
kd-ods, line 111: I thought these and the subsequent lines with 'const' are the ones that are not needed?

kd-ods (changes requested):

tests/data/invalid-statements/expected_errors.csv
kathryn-ods: oops yeah something must have gone wrong with the rebase

kd-ods (approved):

PR #645, opened 2024-04-16
Homepage edits

kathryn-ods:
# Overview

- What does this pull request do? updates the homepage

Closes #

## Translations

## Documentation & Release

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,17 +1,21 @@
-Beneficial Ownership Data Standard (v0.3)
+Beneficial Ownership Data Standard (v0.4)
 ============================================
 
-This website contains the draft specification of the Beneficial Ownership Data Standard (version 0.3) plus support for its use. :doc:`Find out more <about/index>` about the standard's development and how to get involved.
+Information on who owns, controls, and benefits from companies and other corporate vehicles is an essential part of a well-functioning economy and society.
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -42,19 +40,20 @@ This website contains the draft specification of the Beneficial Ownership Data S
 
    .. cssclass:: primaryfrontpageboxlink
 
-      :doc:`Visit the Data Schema <../schema/index>`
+      :doc:`Key Concepts <../schema/concepts>`
 
    .. cssclass:: secondaryfrontpageboxlink
 
-      :doc:`The data model  <../schema/concepts>`
+      :doc:`Modelling requirements  <../schema/modelling-requirements>`
```
odscrachel, line 47: When I build this locally, the link for 'Modelling requirements' and 'system requirements' doesn't show.

kathryn-ods: oops yes I should have mentioned that - these sections don't exist yet. Once the docs are restructured those should work.

kd-ods: @kathryn-ods - I think my preference is to just remove the links entirely for now so that we don't have the build warnings on main. (When we restructure the docs, we can also set up anchors and use :ref: s here instead of :doc:s so that we don't need to put paths in links.)

kathryn-ods: will do!

odscrachel (approved):

PR #651, opened 2024-04-17
moving format inside the bracket on companyFilingsURLs

kathryn-ods:
# Overview

- What does this pull request do? Currently invalid test data for 'companyFilingsURLs is an array of uri format strings' is passing as valid when using a string that is not URI format. When you move format inside items it validates as expected. 

Closes #

## Translations

## Documentation & Release


kd-ods (approved):
Nice spot! Exactly the type of thing that adding test data is meant to flag up.

(In future, feel free to make schema edits on branches where you're adding test data, when issues like this come up. Just flag it in the PR.)

PR #659, opened 2024-04-19
Tests: check for missing files in expected_errors, fixes #657

### LINKED ISSUE #657: invalid test files - extra file names in expected errors doesn't cause an error
if you add an extra line to the expected errors file but that file doesn't exist it doesn't seem to register as an error. Kadie noticed this when I renamed some files but forgot to remove the lines. In the short term we can make sure the number of files = the number of lines in the doc but this may be worth looking at in the future.

rhiaro:
# Overview

Checks for files listed in expected_errors.csv that aren't present on disc

Run the tests, add a stray line to expected_errors.csv or change the spelling of a filename in an existing line, and see failing tests.

Closes #657


kathryn-ods (approved):
Thank you! only note - doesn't pick up duplicated lines. Not sure if it needs to.

PR #662, opened 2024-04-22
Upgrade transifex client

### LINKED ISSUE #428: [i18n] Update transifex client
The transifex client is being deprecated completely in November 2022 in favour of a new one. I just had a poke around at updating now, but right now the new one seems to have some options we need missing from the config, and is reportedly 10x slower than the old one (despite being advertised as faster) so decided to leave it.

Also at the moment part of our process uses sphinx-intl which uses the old transifex client (they have [an issue about upgrading](https://github.com/sphinx-doc/sphinx-intl/issues/60)).

The new client is written in Go rather than Python, so won't be installed with pip as the current one. New installation instructions in the readme would look something like:

> You need to install the transifex commandline client. On Linux and Mac this will most likely work:
> 
```bash
curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
```
> 
> **Restart your shell**, and run `tx --version` to check it installed correctly.
> 
> Please verify this is up to date and also [see other instructions for installation in various environments](https://github.com/transifex/cli#installation).

Just opening this issue to keep an eye on things, and as a reminder as November gets closer.

rhiaro:
# Overview

Updates all dependencies and the transifex config to be compatible with the newest transifex client.

Not much has changed. Transifex client is no longer a python dependency, and now  needs to be installed with a curl command.

Resources on bods-main on transifex are updated to match what is currently in main. For any more schema/codelist/docs changes, the extract and pull steps will need to be run again.

Reinstall dependencies in a new environment, and follow the instructions in BODS Dev Handbook PR https://github.com/openownership/bods-dev-handbook/pull/64

Closes #428

## Translations

## Documentation & Release

- [x] I've checked that the documentation builds without failing. See: https://github.com/openownership/data-standard#building-the-documentation-locally


kd-ods (approved):
I've installed the updated requirements, and I've run the docs build and pytest to check that these updates haven't negatively affected any other processes. So am approving on that basis.

(The .tx/config file is now out of date with what's on main, but we'll update it at the point of freeze.)

PR #708, opened 2024-06-13
Docs tests

rhiaro:
# Overview

Tests for documentation in github workflow

Closes #

## Translations

## Documentation & Release


rhiaro:
@radix0000 Currently the link checker fails on `../_static/statement.json` (and the three others) in schema browser. Do you know what part of the build process these files come from? Or are these links actually incorrect (and should be `../../schema/*.json`)? Wondering if this is a docson thing I haven't managed to figure out yet. But so far it's not obvious to me where these files are supposed to come from.

It also fails for `https://github.com/openownership/data-standard/tree/0.4.0/examples` in examples index, but I'm assuming the latter is because the 0.4 branch doesn't exist yet, but is expected to be.

radix0000:
So seems like there are a couple of issues here:

1) In the docs/_static directory there is a 6 year old broken symlink to the schema directory, and no links for the current schema files. I have created a branch that fixes this: https://github.com/openownership/data-standard/tree/fix_static_links
2) That said the link checker still produces an error because there is one link that is wrong in schema-browser.rst: ../_static/person-statement.json should be ../_static/person-record.json
3) That said these links (except the person-statement.json one) aren't actually broken in the actual docs build, because at some point in build process these files are copied into the docs/_build/dirhtml/_static/ directory, which does raise a question about whether the link checker is fit for purpose if it isn't checking the directory where the files are picked up in the actual built docs, but that is rather a meta point

kd-ods:
> It also fails for https://github.com/openownership/data-standard/tree/0.4.0/examples in examples index, but I'm assuming the latter is because the 0.4 branch doesn't exist yet, but is expected to be.

Yes, that's right.

rhiaro:
> at some point in build process

Well, that's the part I'd like to understand... where does this happen? The linkcheck is part of sphinx, and builds the docs in the process of running itself.. so I'm baffled about what it's missing.

Anyway, I've merged the symlinks fix - thanks @radix0000 

And talked to @kd-ods earlier, and we agreed we can merge this with the remaining broken links - the `person-record` one, and the examples - and they'll be fixed in the staging branch later, when the translations are all finalised.

kd-ods:
@rhiaro - as-is this won't check the build of the translations (and the links in the translated builds), will it? That's probably OK since we need to 'let' the translations get out of sync with the source files on main. Am I thinking about this correctly?

rhiaro:
@kd-ods It generates the `mo` files, but doesn't build the translated html. Could do, if you thought it would be useful? It would catch any typos in links introduced accidentally by translators, for example.

kd-ods (approved):

PR #722, opened 2024-08-09
Annotations required updated

### LINKED ISSUE #721: Annotations required fields
In the schema we have some conditional logic that dictates which fields are required for annotations. In all cases 'statementPointerTarget' and 'motivation' are required. This isn't indicated in the documentation - possibly because they're only indicated as required by the conditional logic? We should probably pull the requirement for these 2 fields out of the conditional logic and then I think it would show in the docs automatically.

kathryn-ods:
# Overview

**What does this pull request do?** fixes #721 

check all the tests pass, build the docs and check the expected fields now show as required 

Closes #

## Translations

## Documentation & Release

### DIFF
--- a/schema/statement.json
+++ b/schema/statement.json
@@ -291,7 +291,8 @@
         }
       },
       "required":[
-        "motivation"
+        "motivation",
+        "statementPointerTarget"
       ],
       "allOf": [
         {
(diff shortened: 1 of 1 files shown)


odscrachel (approved):

kathryn-ods:
@kd-ods I think this ok to merge as won't actually impact the 0.4.0 branch but tagging for a sense check when you're back. Also should we be merging this in to 0.4.0 or 0.4.1 maybe?

PR #727, opened 2024-08-23
Corrected record-id values in example BODS files

### LINKED ISSUE #726: [BUG] - Some example files are incorrectly re-using recordId value for different entities

Some of the files in the examples folder (on [main](https://github.com/openownership/data-standard/tree/main/examples), but also for the 0.4 release) are incorrectly re-using recordId values for different entities. All the example files need to be checked and fixed.

**Suggested resolution**

@radix0000 to identify which files need fixes.

@kd-ods - fix these files on the main branch.

(I suggest that we hold off fixing them on the 0.4 release branch: I suspect that we will uncover other problems with the example data once the new checks are in the Data Review Tool. Eventually we can fix up all the 0.4 example data on main, then apply the fixes to the 0.4 release branch too.)

kd-ods:
# Overview

At least four of the examples files on main (and in the 0.4 release) have some incorrect record ids. See #726

The best way to check the recordIds work is to see that that network graphs look as expected. See the related issue for visuals to test against.

Myles

Closes #726

## Translations

n/a

## Documentation & Release

n/a

### DIFF
--- a/examples/bods-package-annotations.json
+++ b/examples/bods-package-annotations.json
@@ -60,7 +60,7 @@
         "name": "MONSOON HOLDINGS (NO.1) LIMITED"
       }
     },
-    "recordId": "387a14452645",
+    "recordId": "22e8a31863ee",
     "recordStatus": "new",
     "recordType": "entity",
     "recordDetails": {
@@ -101,7 +101,7 @@
       {
         "statementPointerTarget": "/interestedParty",
         "motivation": "identifying",
-        "description": "Company number missing from benenficial ownership statement. Manually reconciled against official register.",
+        "description": "Company number missing from beneficial ownership statement. Manually reconciled against official register.",
         "createdBy": {
           "name": "Jack Lord"
         }
@@ -119,7 +119,7 @@
     "recordType": "relationship",
     "recordDetails": {
       "isComponent": false,
-      "subject": "387a14452645",
+      "subject": "22e8a31863ee",
       "interestedParty": "387a14452645"
     }
   }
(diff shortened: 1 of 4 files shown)


codemacabre:
This all looks good in the visualisation tool except **bods-package-entity-owning-entity.json**, which uses `exclusiveMaximum` for the `maximum` share value. It looks like this is a new key in the `share` object, so I will need to work on making the visualisation tool compatible with that (although it doesn't appear to be documented in the [changelog](https://standard.openownership.org/en/0.4.0/standard/changelog.html) so I'm adding this comment here to check that it is correct before approving).

kd-ods:
Ah, yes. The changlelog has the teaser entry "Interest.share properties and requirements updated. Exact values and ranges are now represented in simpler ways."

To be precise: `exclusiveMinimum` and `exclusiveMaximum` properties did exist before in 0.3 etc, but they were booleans and acted as modifiers for the meaning of the `minimum` and `maximum` properties. From 0.4, `exclusiveMinimum` and `exclusiveMaximum` are now numbers. If publishers need to show that a share interest lies within a range, they use one of `minimum` and `exclusiveMinimum` and then one of `maximum` and `exclusiveMaximum`. (Do you think we should open up a visualiser issue where we can work on specifying the visualiser handling of the share object? It's not a simple matter, when the tool shouldn't get into the business of only rendering valid data.)

codemacabre:
> (Do you think we should open up a visualiser issue where we can work on specifying the visualiser handling of the share object? It's not a simple matter, when the tool shouldn't get into the business of only rendering valid data.)

I don't think it's necessary at this stage - the fix is a fairly straightforward conditional to check for (`minimum` or `exclusiveMinimum`) and (`maximum` or `exclusiveMaximum`) which should make valid data render correctly.

codemacabre (approved):

PR #735, opened 2024-11-21
Example drt main

### LINKED ISSUE #738: [BUG] - Example files need to be corrected on main and 0.4 release

We effectively hand-rolled the example data for the 0.4 release. Unsurprisingly, the use of the example files for testing the updates to the Data Review Tool and the Visualiser uncovered various issues with the data. 

**Suggested resolution**

We should correct the example files. Once the tools are released we can do final tests on the corrected example files, merge into main, and then replicate on the release branch.

kathryn-ods:
# Overview

Makes minor fixes to the examples files, identified by the DRT

Run a few through the DRT to check they're fine 

Closes #738 

## Translations

## Documentation & Release

### DIFF
--- a/docs/standard/modelling/identifiers.rst
+++ b/docs/standard/modelling/identifiers.rst
@@ -80,7 +80,7 @@ See this :ref:`example data <examples-plc>` for a valid use of MICs.
 
 Two properties in an Entity Statement’s :ref:`Securities Listing <schema-securities-listing>` object identify where a security is traded: the ``operatingMarketIdentifierCode`` of the operating market and the ``marketIdentifierCode``. The ``marketIdentifierCode`` SHOULD be the same as the ``operatingMarketIdentifierCode`` if the security is traded on a main exchange. The ``marketIdentifierCode`` will differ where a security is traded on a segment of an exchange.
 
-MICs are standardised, issued and maintained by SWIFT as ISO 10383. The specification of ``operatingMarketIdentifierCode`` and ``marketIdentifierCode`` is part of the ISO standard.
+MICs are standardised, issued and maintained by SWIFT as `ISO 10383 <https://www.iso20022.org/market-identifier-codes>`_. The specification of ``operatingMarketIdentifierCode`` and ``marketIdentifierCode`` is part of the ISO standard.
 
 .. guidance-identifiers-securities:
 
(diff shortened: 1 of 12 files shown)


kathryn-ods:
@kd-ods in addition to the fixes here all examples had the error 

> beneficialOwnershipOrControl expected to be 'true' in at least one Relationship statement with a person as an interested party. If this dataset contains beneficial owners, check that beneficialOwnershipOrControl is correctly used.	

Do we need to fix this?

kathryn-ods:
I'm not sure why the docs linkcheck keeps breaking - I haven't made any docs changes so it shouldn't be. I think there's some sort of issue with the way it interacts with the ISO site.

kd-ods:
> @kd-ods in addition to the fixes here all examples had the error
> 
> > beneficialOwnershipOrControl expected to be 'true' in at least one Relationship statement with a person as an interested party. If this dataset contains beneficial owners, check that beneficialOwnershipOrControl is correctly used.
> 
> Do we need to fix this?

I don't think we need to change the example files.

(It's a bit of an anomalous type of error message... since it's not an error, as such. It's a prompt for users of the tool to check that they are correctly using that field IF their dataset contains beneficial owners. There are definitely improvements to the user interface of the Data Review Tool which could make that clearer.)

kd-ods:
The commit above fixes an issue [spotted here](https://github.com/openownership/visualisation-tool/issues/176#issuecomment-2505828115).

kathryn-ods:
@kd-ods I think this still needs merging in? The link check is breaking for some reason other than the docs itself

kd-ods:
It looks like the particular link that is failing the check is https://www.iso20022.org/market-identifier-codes - it's apparently timing out. However the link does not seem to be broken if checked by hand. 

I suspect that bots checking links are not welcomed in iso-land.

Do you think we should wait until the data review tool is released, do a final run of the data through it, before merging @kathryn-ods ?

kathryn-ods:
> It looks like the particular link that is failing the check is https://www.iso20022.org/market-identifier-codes - it's apparently timing out. However the link does not seem to be broken if checked by hand.

I asked @rhiaro to look at this and they said

> looks like there's something wrong with the ssl cert handshake and it times out. Not sure how to fix it off the top of my head. I think it's an issue with openssl on ubuntu .. I haven't looked into exactly what is running the tests on github actions, but there's probably a version upgrade somewhere in there that would resolve it


> Do you think we should wait until the data review tool is released, do a final run of the data through it, before merging

I don't have strong feelings either way - what do you think?

kd-ods (changes requested):
A couple of the files have discrepancies in the statementDates which need to be fixed.

kd-ods (approved):

PR #758, opened 2026-04-23
Add json-pointer format to statementPointerTarget

### LINKED ISSUE #707: Add string format:"json-pointer" to statementPointerTarget property
**Summary**

To validate the basic format of the `statementPointerTarget` property in the Annotations object, we should add the string format: "json-pointer".

**Additional context and details**

The property is currently defined in the Statement.json file as:

```
        "statementPointerTarget": {
          "title": "Statement Fragment Pointer",
          "description": "An RFC6901 JSON Pointer (https://tools.ietf.org/html/rfc6901) describing the target fragment of the statement that this Annotation applies to, starting from the root of the Statement. An empty string (\"\") indicates that the Annotation applies to the whole Statement.",
          "type": "string"
        }
```

StephenAbbott:
## Summary

Fixes #707.

The `statementPointerTarget` property on the `Annotation` object is described as an RFC6901 JSON Pointer but the schema only constrained it to `type: string`. This adds `"format": "json-pointer"` so validators reject values that are not well-formed pointers (e.g. missing leading `/` or containing invalid `~` escapes).

Change in `schema/statement.json`:

```diff
         "statementPointerTarget": {
           "title": "Statement Fragment Pointer",
           "description": "An RFC6901 JSON Pointer ...",
-          "type": "string"
+          "type": "string",
+          "format": "json-pointer"
         },
```

## Test plan

- [x] Added `tests/data/invalid-statements/statement_annotations_statementPointerTarget_format.json` with an invalid pointer (missing leading `/`)
- [x] Added matching row to `tests/data/invalid-statements/expected_errors.csv`
- [x] `pytest tests/test_data.py` — new test passes; no new failures (pre-existing 12 format-related failures on `main` in the local sandbox are environmental and unchanged here)
- [x] All existing valid fixtures use well-formed pointers (`""`, `/recordDetails/...`, `/statementDate`) so none break under the new constraint

https://claude.ai/code/session_019ksAaAodXk1wNCLcuEVkXC

---
_Generated by [Claude Code](https://claude.ai/code/session_019ksAaAodXk1wNCLcuEVkXC)_

### DIFF
--- a/schema/statement.json
+++ b/schema/statement.json
@@ -226,7 +226,8 @@
         "statementPointerTarget": {
           "title": "Statement Fragment Pointer",
           "description": "An RFC6901 JSON Pointer (https://tools.ietf.org/html/rfc6901) describing the target fragment of the statement that this Annotation applies to, starting from the root of the Statement. An empty string (\"\") indicates that the Annotation applies to the whole Statement.",
-          "type": "string"
+          "type": "string",
+          "format": "json-pointer"
         },
         "creationDate": {
           "title": "Creation Date",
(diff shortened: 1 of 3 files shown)


kd-ods:
Thanks, @StephenAbbott . This edit will require some work our end so I'll tag it as a good PR to start with if there's a future round of development on the schema.

